PR #4, opened 2020-09-06
Fixed capability type

guillemcordoba:
As new the new holochain conductor interface now checks for the type and length of the capability secret, but not its contents, the holochain-conductor-api types for that parameter in the `callZome` function should align to that. Otherwise, right now a consumer of either this library or tryorama has to hack their `node_modules` folder for the `callZomes` to work.

Any suggestions better than `Buffer` are welcomed, this is more of a temporary fix until good capabilities are implemented.

See also https://github.com/Holo-Host/tryorama-rsm/pull/2.


maackle:
Thank you @guillemcordoba!

PR #8, opened 2020-09-25
Switch to @msgpack/msgpack

guillemcordoba:
Completed the switch to [msgpack-javascript](https://www.npmjs.com/package/@msgpack/msgpack).

Also I added `npm run format` to format all source code files since I saw some styles inconsistencies. @maackle if you don't want this I can take it out of this PR.

### DIFF
--- a/src/websocket/app.ts
+++ b/src/websocket/app.ts
@@ -15,7 +15,7 @@
  *        console.error('problem installing DNA:', err)
  *      })
  */
-import msgpack from '@msgpack/msgpack';
+import * as msgpack from '@msgpack/msgpack';
 
 import { AppApi, CallZomeRequest, CallZomeResponse, AppInfoRequest, AppInfoResponse, CallZomeRequestGeneric, CallZomeResponseGeneric } from '../api/app'
 import { WsClient } from './client'
--- a/src/websocket/app.ts
+++ b/src/websocket/app.ts
@@ -15,7 +15,7 @@
  *        console.error('problem installing DNA:', err)
  *      })
  */
-import * as msgpack from 'msgpack-lite'
+import msgpack from '@msgpack/msgpack';
 
 import { AppApi, CallZomeRequest, CallZomeResponse, AppInfoRequest, AppInfoResponse, CallZomeRequestGeneric, CallZomeResponseGeneric } from '../api/app'
 import { WsClient } from './client'
--- a/src/websocket/client.ts
+++ b/src/websocket/client.ts
@@ -1,5 +1,5 @@
 import Websocket from 'isomorphic-ws'
-import msgpack from '@msgpack/msgpack';
+import * as msgpack from '@msgpack/msgpack';
 import { nanoid } from 'nanoid'
 
 /**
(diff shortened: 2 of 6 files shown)


src/websocket/app.ts
```diff
@@ -15,43 +15,64 @@
...
+const callZomeTransform: Transformer<
+  CallZomeRequestGeneric<any>,
+  CallZomeRequestGeneric<Buffer>,
+  CallZomeResponseGeneric<Buffer>,
+  CallZomeResponseGeneric<any>
+> = {
   input: (req: CallZomeRequestGeneric<any>): CallZomeRequestGeneric<Buffer> => {
-    req.payload = msgpack.encode(req.payload)
-    return req
+    req.payload = msgpack.encode(req.payload);
```
Connoropolous: so the api was identical? 
and this is the core change?

src/websocket/client.ts
```diff
@@ -9,69 +9,75 @@ import { nanoid } from 'nanoid'
...
         socket.onmessage = (encodedMsg: any) => {
-          const msg = msgpack.decode(encodedMsg.data)
-          if (signalCb && msg.type === 'Signal') {
-            signalCb(msgpack.decode(msg.data))
-          } else if (msg.type === 'Response') {
-            const id = msg.id
-            if (hw.pendingRequests[id]) {
-              // resolve response
-              hw.pendingRequests[id].fulfill(msgpack.decode(msg.data))
+          encodedMsg.data.arrayBuffer().then((data: any) => {
```
Connoropolous: is there a type that we can refer to for `encodedMsg`, other than `any`?

was the change to using `encodedMsg.data.arrayBuffer()` necessary as part of the change to using `msgpack/msgpack` ? 

this would be the other core change I think?

src/websocket/client.ts
```diff
@@ -9,69 +9,75 @@ import { nanoid } from 'nanoid'
...
+            const msg: any = msgpack.decode(data);
```
Connoropolous: why does `encodedMsg` need to be decoded twice?

and can the two secondary calls to `msgpack.decode(msg.data)` be consolidated?

Connoropolous (commented):
I know the team is busy and spread between many things... happy to help where I can as well, so just let me know @guillemcordoba if you want some reviews on PRs too

This has me thinking, with curiosity, what kind of social protocol for approval do these repos have?

src/websocket/app.ts
guillemcordoba: Api is identical, yes.

src/websocket/client.ts
guillemcordoba: Yes exactly, @msgpack/msgpack does not accept a Blob, only arrayBuffers. Don't know about the any, I think ts cannot assume anything else.

src/websocket/client.ts
guillemcordoba: Hum good question, @maackle ? I think it's because there are two serializations: 1) the message itself, 2) the payload inside the message.

maackle: Yep exactly, there are two layers of serialization, one to cross over the Wasm boundary, and another to cross over the Websocket

maackle (approved):
This looks good, thank you for doing this. I need to be better about checking GH notifications daily.

I trust that the tests passed. (We need to get CI hooked up for this and other repos!)

maackle (changes requested):
Oops, spoke too soon. I decided to run the tests, but there is a syntax error, probably due to the change in module resolution settings. I don't understand that change, so could you just make sure you can successfully `npm install && npm test` ? Unless the tests are broken for some other reason, it's been a while since I've run them..

guillemcordoba:
Oh yeah sorry, yesterday I meant to run them before notifying you but with all the compilation steps I got lost into other stuff. So yeah absolutely will run them. I should have posted here.

guillemcordoba:
Actually @maackle I am having trouble with the instructions for the tests. No `wasm32-unknown-unknown` folder is generated on the build step. Do you know if this has changed recently @thedavidmeister?

src/websocket/client.ts
Connoropolous: oh ok. I follow. We can drop that in as a comment here in this code, to add a little more clarity on that piece

maackle:
@guillemcordoba whoops, yes, the build process did change, and now the test wasms should be available in `.wasm_target/wasm32-unknown-unknown/release` of the holochain repo dir

maackle:
updated README

guillemcordoba:
Okey, tests are passing now. The reason that they weren't passing before were due to the `encodedMsg.data` being a `Buffer` instead of a `Blob`, which is what I found in the browser... For now I've added a check to support both `Buffer` and `Blob` types, but I don't know if we should we look more into this or not. @maackle ?

maackle:
Thank you @guillemcordoba! I confirmed that tests pass. Given that, I think this is OK to merge. In the future we should have browser tests as well, but the tests at least confirm that it works in node.

maackle (approved):

Connoropolous:
I guess you're talking about this because holochain-conductor-api is supposed to work isometrically in the browser and in node?

I am trying to imagine why it would be a buffer vs a blob depending on which of those its running in. Doesn't that depend on where a message is sent from, not where it arrives?

PR #15, opened 2020-10-15
CI

### LINKED ISSUE #14: ci

Connoropolous:
closes #14 

`nix-shell --run conductor-api-test`

This also has the benefit of being explicit about at which commit (since there's no RSM releases/tags yet) of holochain/holochain this code is compatible with. 

Tests do pass, just need @thedavidmeister or @maackle  to enable CircleCi for this repo :)


config.nix
```diff
@@ -0,0 +1,101 @@
...
+  # the release process needs to add some commits for changelog etc.
+  commit = "926efebf26d69ee4d6cb9ac2d78fe04531e1458b";
+
+  # the semver for prev and current releases
+  # the previous version will be scanned/bumped by release scripts
+  # the current version is what the release scripts bump *to*
+  version = {
+   current = "0.0.1-dev.9";
+   previous = "_";
+  };
```
maackle, line 72: Hmm I really don't want to involve nix in the simple process of cutting a new npm release. Raw npm is perfectly capable of this. @thedavidmeister we talked about using holonix less for ancillary repos like this, would this be a good time to break from the mold?

Connoropolous: Ah, I'm not opinionated on how releases should happen. Can leave that part blank, and just not use `hn-release-github` etc.

Connoropolous:
I have heard from David Meister I am going to switch to Github Actions

Connoropolous:
closing in favor of #16

PR #16, opened 2020-10-16
CI and extended tests

Connoropolous:
This is no longer CI with nix-shell, this is CI with Github Actions (node, rust), and a couple bash scripts

fixes #12 and #3 and adds the test cases for them

CI looks like it's going to take a really long time to compile both `holochain` and `dna-util`. About 25 minutes. 
That's pretty painful. At least we can cut that down by pre-packing the DNA as a `.dna.gz` committed to git, and avoid compiling `dna-util`. 

It's a tradeoff, I preferred to allow for full compilation on CI, so that you don't forget locally and CI fails for the wrong reasons. 
Thoughts?

PR #17, opened 2020-10-16
Return legible signal data to ui

JettTech:
### Updates:
 - decode serialized signal payload
 - return uniform call format to ui (ie: { type, data })


Connoropolous (approved):
Good to see this case properly handled. 
I hadn't tested this part of the system yet myself, but had a feeling it wasn't resolved. 

This PR makes me think that this repo really needs a default code formatter, because we each are introducing various syntax differences, which usually makes it harder to read PRs too.

P.s. if you haven't heard there is 'hApp devs with early access' team that's been created that I'm on, which is why I can comment and feedback on PRs.

zo-el (approved):
LGTM

Connoropolous:
@JettTech CI is now working over in `ci-with-new-tests-and-fixes` #16 if you wanted to change the `base` of this PR, run CI, and have a more convenient way to run the tests. Not necessary though

PR #25, opened 2020-10-27
Check if buffer is defined

guillemcordoba:
This allows for library use without having to include the feross `buffer` library.

### DIFF
--- a/src/websocket/client.ts
+++ b/src/websocket/client.ts
@@ -72,8 +72,8 @@ export class WsClient {
         socket.onmessage = async (encodedMsg: any) => {
           let data = encodedMsg.data
 
-          // If data is not a buffer, it will be a blob
-          if (!Buffer.isBuffer(data)) {
+          // If data is not a buffer (nodejs), it will be a blob (browser)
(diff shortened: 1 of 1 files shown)


src/websocket/client.ts
```diff
@@ -72,8 +72,8 @@ export class WsClient {
         socket.onmessage = async (encodedMsg: any) => {
           let data = encodedMsg.data
 
-          // If data is not a buffer, it will be a blob
-          if (!Buffer.isBuffer(data)) {
+          // If data is not a buffer (nodejs), it will be a blob (browser)
+          if (typeof Buffer === "undefined" || !Buffer.isBuffer(data)) {
```
Connoropolous, line 76: is there any need for this part anymore `!Buffer.isBuffer(data)`? was it just covering the browser case?

guillemcordoba: Yeah that covers the case in which the web app injects the Buffer library. But since we are in the browser, the data will still be a blob.

Connoropolous (approved):

PR #27, opened 2020-10-28
bumped holochain to 96815c179373bbc1ffca84c48e97c5a7746e4ead (Oct 27)

zo-el:


Connoropolous (approved):
🐈  purrfect. I was going to do this this morning, instead I wake up and it's done :)
it all looks right.

Connoropolous (changes requested):
oops, ok wait. CI is failing. 
That's because of a test that checks the behavior of ZomeCallUnauthorized. Let me fix that one.

test/e2e/fixture/zomes/foo/Cargo.lock
```diff
@@ -1,511 +0,0 @@
-# This file is automatically @generated by Cargo.
```
Connoropolous, line 1: @zo-el why delete Cargo.lock?

Connoropolous: Oh, I get it, because there's a top-level one now

zo-el: That was an oversight.

Connoropolous (approved):
we've got CI passing now so re-approved

PR #30, opened 2020-11-10
Update for HoloHash serialization

Connoropolous:
- [x] change `Hash` to `HoloHash` and make it a simple Buffer, instead of an object/struct
- [x] add P2P_BOOTSTRAP_URL="" necessary to prevent fatal errors
- [x] update holochain version to https://github.com/holochain/holochain/commit/beaab9949fbff121ffd138f36c4e61c5693b8fd6

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -1,8 +1,8 @@
-export type Hash = Buffer // length 39
-export type AgentPubKey = Hash
+export type HoloHash = Buffer // length 39
+export type AgentPubKey = HoloHash
 export type AppId = string
 export type CapSecret = Buffer
-export type CellId = [Hash, AgentPubKey]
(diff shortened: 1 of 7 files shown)


src/api/types.ts
```diff
@@ -1,26 +1,22 @@
...
 export type CapSecret = Buffer
-export type CellId = [Hash, AgentPubKey]
+export type CellId = [HoloHash, AgentPubKey]
 export type CellNick = string
 export type DnaProperties = any
 export type InstalledApp = {
-  app_id: AppId,
-  cell_data: Array<InstalledCell>,
+  app_id: AppId
+  cell_data: Array<InstalledCell>
```
maackle, line 10: whoa, this works without commas?

maackle (approved):
Nice :+1:

PR #36, opened 2020-11-26
Socket open onError fixes

zippy:
- [x] Websocket error results (on browser) do not actually return the error code.  
- [x] calling send on a closed websocket was causing promises never returned errors.

### DIFF
--- a/src/websocket/client.ts
+++ b/src/websocket/client.ts
@@ -62,7 +62,7 @@ export class WsClient {
       // with uncaught exception
       socket.onerror = (e) => {
         reject(new Error(
-          `error while connecting to holochain at ${url}: ${e.error.name}, ${e.error.message}`
+          `could not connect to holochain conductor, please check that a conductor service is running and available at ${url}`
         ))
       }
       socket.onopen = () => {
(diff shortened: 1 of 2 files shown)


Connoropolous (approved):
Makes sense, and are very simple changes 👍

src/websocket/client.ts
```diff
@@ -57,15 +61,9 @@ export class WsClient {
...
-          reject(
-            new Error(
-              `could not connect to holochain conductor, please check that a conductor service is running and available at ${url}`
-            )
-          )
-        } else {
-          reject(e)
-        }
+        reject(new Error(
+          `could not connect to holochain conductor, please check that a conductor service is running and available at ${url}`
```
peeech, line 65:
```suggestion
          `error while connecting to holochain at ${url}: ${e.error.name}, ${e.error.message}`
```
Would it be useful to pass an error message to the client? I am not sure.

zippy: It would be nice, but I think this is what caused the error in the first place.  The error object does not seem to exist in the browser context.  Maybe it does in the node context, but this code throws `Cannot read property 'name' of undefined` so I removed it.

peeech: ha, interesting

peeech (approved):

PR #39, opened 2020-12-01
Update api to include agent inject feature

zippy:
The PR adds the agent injection that was added to the admin API in [holochain/#501](https://github.com/holochain/holochain/pull/501)

### DIFF
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -32,6 +32,11 @@ export type ListCellIdsResponse = Array<CellId>
 export type ListActiveAppsRequest = void
 export type ListActiveAppsResponse = Array<InstalledAppId>
 
+export type AgentInfoSigned = any
+
+export type RequestAgentInfoRequest = { cell_id: CellId|null }
+export type RequestAgentInfoResponse = Array<AgentInfoSigned>
+
 export interface AdminApi {
   activateApp: Requester<ActivateAppRequest, ActivateAppResponse>
   attachAppInterface: Requester<AttachAppInterfaceRequest, AttachAppInterfaceResponse>
@@ -42,6 +47,7 @@ export interface AdminApi {
   listDnas: Requester<ListDnasRequest, ListDnasResponse>
   listCellIds: Requester<ListCellIdsRequest, ListCellIdsResponse>
   listActiveApps: Requester<ListActiveAppsRequest, ListActiveAppsResponse>
+  requestAgentInfo: Requester<RequestAgentInfoRequest, RequestAgentInfoResponse>
 }
 
 
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -36,6 +36,8 @@ export type AgentInfoSigned = any
 
 export type RequestAgentInfoRequest = { cell_id: CellId|null }
 export type RequestAgentInfoResponse = Array<AgentInfoSigned>
+export type AddAgentInfoRequest = { agent_infos: Array<AgentInfoSigned> }
+export type AddAgentInfoResponse = any
 
 export interface AdminApi {
   activateApp: Requester<ActivateAppRequest, ActivateAppResponse>
@@ -48,6 +50,7 @@ export interface AdminApi {
   listCellIds: Requester<ListCellIdsRequest, ListCellIdsResponse>
   listActiveApps: Requester<ListActiveAppsRequest, ListActiveAppsResponse>
   requestAgentInfo: Requester<RequestAgentInfoRequest, RequestAgentInfoResponse>
+  addAgentInfo: Requester<AddAgentInfoRequest, AddAgentInfoResponse>
 }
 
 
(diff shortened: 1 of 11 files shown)


test/e2e/index.ts
```diff
@@ -136,3 +136,63 @@ test('error is catchable when holochain socket is unavailable', async (t) => {
...
+                    path: `${FIXTURE_PATH}/test.dna.gz`,
+                    nick,
+                }
+            ]
+        })
+        t.ok(result)
+        const app2_cell  = result.cell_data[0][0]
+        await admin2.activateApp({ installed_app_id })
+
+        const conductor2_agentInfos = await admin2.requestAgentInfo({cell_id: null});
```
Connoropolous: What does it mean to pass cell_id: null?
Unfiltered? All peers?

Connoropolous (commented):
Just an initial scan...

test/e2e/index.ts
Connoropolous: It'd be really good to just have some simple comments around the different phases of this test

Connoropolous: It'd be really good to just have some simple comments around the different phases of this test

Connoropolous:
I really like that this is possible now because not only is it good for the test framework,  but it will be good for other third party peer discovery solutions and options.

test/e2e/index.ts
philipbeadle: yes

Connoropolous:
@philipbeadle did the tests pass locally? 
CI is failing with:
```
   can inject agents
    ✔  should be truthy
    ✔  should be truthy
    ✔  should be truthy
    ✖  [object Object] at undefined
        undefinedundefined
        Conductor exited with code null
        Conductor exited with code null
```

src/api/admin.ts
```diff
@@ -32,6 +32,13 @@ export type ListCellIdsResponse = Array<CellId>
 export type ListActiveAppsRequest = void
 export type ListActiveAppsResponse = Array<InstalledAppId>
 
+export type AgentInfoSigned = any
```
Connoropolous, line 36: instead of `any` here, why not `console.log` an instance of one of these and map out the type here?

Connoropolous (changes requested):

src/api/admin.ts
zippy: done

test/e2e/index.ts
zippy: done

src/api/admin.ts
```diff
@@ -32,6 +32,17 @@ export type ListCellIdsResponse = Array<CellId>
 export type ListActiveAppsRequest = void
 export type ListActiveAppsResponse = Array<InstalledAppId>
 
+export type AgentInfoSigned = {
+    agent: AgentPubKey,
```
freesig: Ok this is wrong. It's not an `AgentPubKey`.
The `AgentInfoSigned` type should never be used at this layer.
There's no need to deserialize the type at all, just take the array of bytes from the websocket and pass it to add agent info

src/api/admin.ts
```diff
@@ -32,6 +32,17 @@ export type ListCellIdsResponse = Array<CellId>
...
+    signature: any, // don't yet have a type for signatures
+    agent_info: any, // don't yet have a type for agent_info
+}
+
+export type RequestAgentInfoRequest = { cell_id: CellId|null }
+export type RequestAgentInfoResponse = Array<AgentInfoSigned>
```
freesig, line 44:
```suggestion
export type RequestAgentInfoResponse = Buffer
```
I think this is what you want but not 100% sure. Whatever the equivalent of `Vec<u8>` is.

src/api/admin.ts
```diff
@@ -32,6 +32,17 @@ export type ListCellIdsResponse = Array<CellId>
...
+export type AddAgentInfoRequest = { agent_infos: Array<AgentInfoSigned> }
```
freesig, line 45:
```suggestion
export type AddAgentInfoRequest = { agent_infos: Buffer }
```

freesig (changes requested):
Kepp the agent info type opaque because it's not using holochaib types like `AgentPubKey` it's using kitsune type which you don't have here

src/api/admin.ts
zippy: Hmmm.  To not deserialize the data from a message requires changing the plumbing quite a bit.  The code is generalized to deserialize every message type.  But even if this is "wrong" as an interpretation of the data when deserialized, I think it is still working because in the test when it is reserialized, and passed into AddAgentInfo, and then pulled back out and compared with the original data it all matches.  AgentPubKey happens to be just a Buffer. So even though we don't "need" to serialize and reserialize, that doesn't seem to be a problem.  Just incase I changed the type to "any" and added the deepequal test.

src/api/admin.ts
zippy: see above

src/api/admin.ts
zippy: ditto

Connoropolous (approved):

PR #41, opened 2020-12-10
set timeout

zo-el:
- [x] update websocket to reject promise of sends after timeout
- [x] ability to set timeout on a per-call basis
- [x] ability to set default timeout on per Websocket instances.
- [x] update readme with instructions and current version of holochain tested against

### DIFF
--- a/src/websocket/admin.ts
+++ b/src/websocket/admin.ts
@@ -17,16 +17,16 @@
 
 import * as Api from '../api/admin'
 import { WsClient } from './client'
-import { catchError, promiseTimeout } from './common'
+import { catchError, promiseTimeout, DEFAULT_TIMEOUT } from './common'
 import { Transformer, requesterTransformer, Requester } from '../api/common'
 
 export class AdminWebsocket implements Api.AdminApi {
   client: WsClient
-  timeout?: number
+  timeout: number
 
   constructor(client: WsClient, timeout?: number) {
     this.client = client
-    this.timeout = timeout
+    this.timeout = timeout === undefined ? DEFAULT_TIMEOUT : timeout
   }
 
   static async connect(url: string, timeout?: number): Promise<AdminWebsocket> {
--- a/src/websocket/admin.ts
+++ b/src/websocket/admin.ts
@@ -34,7 +34,7 @@ export class AdminWebsocket implements Api.AdminApi {
 
   _requester = <ReqO, ReqI, ResI, ResO>(tag: string, transformer?: Transformer<ReqO, ReqI, ResI, ResO>) =>
     requesterTransformer(
-      (req, timeout) => promiseTimeout(this.client.request(req), timeout).then(catchError),
+      (req, timeout) => promiseTimeout(this.client.request(req), tag, timeout).then(catchError),
       tag,
       transformer
     )
--- a/src/websocket/admin.ts
+++ b/src/websocket/admin.ts
@@ -22,21 +22,21 @@ import { Transformer, requesterTransformer, Requester } from '../api/common'
 
 export class AdminWebsocket implements Api.AdminApi {
   client: WsClient
-  timeout: number
+  defaultTimeout: number
 
-  constructor(client: WsClient, timeout?: number) {
+  constructor(client: WsClient, defaultTimeout?: number) {
     this.client = client
-    this.timeout = timeout === undefined ? DEFAULT_TIMEOUT : timeout
+    this.defaultTimeout = defaultTimeout === undefined ? DEFAULT_TIMEOUT : defaultTimeout
   }
 
-  static async connect(url: string, timeout?: number): Promise<AdminWebsocket> {
+  static async connect(url: string, defaultTimeout?: number): Promise<AdminWebsocket> {
     const wsClient = await WsClient.connect(url)
-    return new AdminWebsocket(wsClient, timeout)
+    return new AdminWebsocket(wsClient, defaultTimeout)
   }
 
   _requester = <ReqO, ReqI, ResI, ResO>(tag: string, transformer?: Transformer<ReqO, ReqI, ResI, ResO>) =>
     requesterTransformer(
-      (req, timeout) => promiseTimeout(this.client.request(req), tag, timeout || this.timeout).then(catchError),
+      (req, timeout) => promiseTimeout(this.client.request(req), tag, timeout || this.defaultTimeout).then(catchError),
       tag,
       transformer
     )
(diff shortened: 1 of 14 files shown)


Connoropolous (commented):
This doesn't really look like it will work because this.socket.send is not a promise, and wasn't being returned, and Promise.race is not being returned either. You would have to perform those calls inside the new Promise block with the fulfill that is actually getting returned

src/websocket/admin.ts
```diff
@@ -34,7 +34,7 @@ export class AdminWebsocket implements Api.AdminApi {
 
   _requester = <ReqO, ReqI, ResI, ResO>(tag: string, transformer?: Transformer<ReqO, ReqI, ResI, ResO>) =>
     requesterTransformer(
-      req => this.client.request(req).then(catchError),
+      req => promiseTimeout(15000, this.client.request(req)).then(catchError),
```
Connoropolous:
```suggestion
      req => promiseTimeout(15000, this.client.request(req).then(catchError)),
```

src/websocket/app.ts
```diff
@@ -36,7 +36,7 @@ export class AppWebsocket implements AppApi {
 
   _requester = <ReqO, ReqI, ResI, ResO>(tag: string, transformer?: Transformer<ReqO, ReqI, ResI, ResO>) =>
     requesterTransformer(
-      req => this.client.request(req).then(catchError),
+      req => promiseTimeout(15000, this.client.request(req)).then(catchError),
```
Connoropolous:
```suggestion
      req => promiseTimeout(15000, this.client.request(req).then(catchError)),
```

src/websocket/common.ts
```diff
@@ -5,3 +5,26 @@ export const catchError = (res: any) => {
...
+  return new Promise((res, rej) => {
+    Promise.race([
+      promise,
+      timeout
+    ]).then((a) => {
+      clearTimeout(id);
+      return res(a)
+    })
+    .catch(e => {
+      return rej(new Error(e))
```
Connoropolous:
```suggestion
      return rej(e)
```
I think doing this error forwarding you want to forward it plain, not re-wrapped in an Error since it already should be

Connoropolous (commented):
I'd recommend keeping the `catchError` limited to its intended catch scope. This is a much better looking solution that should work though. perhaps timeout should be configurable per-call but I'm not concerned about that right now

zippy (approved):
LGTM

src/websocket/admin.ts
```diff
@@ -34,7 +34,7 @@ export class AdminWebsocket implements Api.AdminApi {
 
   _requester = <ReqO, ReqI, ResI, ResO>(tag: string, transformer?: Transformer<ReqO, ReqI, ResI, ResO>) =>
     requesterTransformer(
-      req => this.client.request(req).then(catchError),
+      (req, timeout) => promiseTimeout(this.client.request(req), timeout).then(catchError),
```
Connoropolous:
```suggestion
      (req, timeout) => promiseTimeout(this.client.request(req).then(catchError), timeout),
```

src/websocket/app.ts
```diff
@@ -36,7 +36,7 @@ export class AppWebsocket implements AppApi {
 
   _requester = <ReqO, ReqI, ResI, ResO>(tag: string, transformer?: Transformer<ReqO, ReqI, ResI, ResO>) =>
     requesterTransformer(
-      req => this.client.request(req).then(catchError),
+      (req, timeout) => promiseTimeout(this.client.request(req), timeout).then(catchError),
```
Connoropolous:
```suggestion
      (req, timeout) => promiseTimeout(this.client.request(req).then(catchError), timeout),
```

Connoropolous (dismissed):
catchError is only really intended for catching what comes up from this.client.req

src/websocket/admin.ts
zo-el: These changes will have some Typescript errors. 
Also If the `Promise.race` throws an error it will wait for the other condition it complete (i.e. the timeout) which we do not want.

Also, I don't see the catch scope as very limited because it also handles errors when we are not connected to the Websocket. So I think it is able to handle the timeout error should not be an issue.
What do you think?

src/websocket/app.ts
maackle: I don't remember what `tag` is here, but if it has helpful context you could pass it into `promiseTimeout` to include in the timeout error message

src/websocket/common.ts
```diff
@@ -1,7 +1,33 @@
...
 }
+
+export const promiseTimeout = (promise: Promise<any>, ms?: number) => {
+  let id
+  let time = ms === undefined ? DEFAULT_TIMEOUT : ms
+
+  let timeout = new Promise((resolve, reject) => {
+    id = setTimeout(() => {
+      clearTimeout(id);
+      reject(new Error('Timed out in '+ time + 'ms.'))
```
maackle: could use `tag` here (if it makes sense to do so)
```suggestion
      reject(new Error(`Timed out in ${time}ms: ${tag}'))
```

src/websocket/admin.ts
maackle: It's OK if the timeout `reject` gets called after the main promise is resolved, it just won't do anything after the overall race promise is resolved via the main promise

maackle (approved):
Looks good, with suggestions

src/websocket/admin.ts
Connoropolous: I will trust Michaels review of it, he's the original author

zippy:
In thinking about this, I'm wondering if we shouldn't be dynamically setting the default timeout in the WsClient object when it's constructed.  That way users can set different default timeouts for different use-cases, i.e. especially between zomeCall and admin use cases which really probably want different timeouts.  @Connoropolous @maackle @zo-el ?

zippy (changes requested):
Need to update readme with timeout examples, and also test against latest holochain version

src/websocket/admin.ts
```diff
@@ -22,19 +22,21 @@ import { Transformer, requesterTransformer, Requester } from '../api/common'
 
 export class AdminWebsocket implements Api.AdminApi {
   client: WsClient
+  timeout?: number
```
zippy: can we call this `defaultTmeout` just to be super clear?

src/websocket/app.ts
```diff
@@ -24,19 +24,21 @@ import { Transformer, requesterTransformer, Requester } from '../api/common'
 
 export class AppWebsocket implements AppApi {
   client: WsClient
+  timeout?: number
```
zippy: ditto defaultTmeout.

zippy (approved):
Just small wording change, and we still need to update the README.

PR #43, opened 2021-01-06
bug fix for use in ts UI

zo-el:
Hi All,

I pulled this pkg into a UI that was built in typescript and it was not able to build this pkg.

This is the error it was throwing 
```
"isomorphic-ws/index"' has no default export
```

This small change fixes the issue.


maackle (approved):
The diff is hard to read for some reason, there are a bunch of failed checks, and yet the tests pass. As long as this works, I'm ok with it!

PR #44, opened 2021-01-09
Add support for register-dna

zippy:
Holochain's ConductorAPI now includes a new function "RegisterDNA" that allows separating loading DNAs into holochains wasm database, from installing happs.  This function allows you to specify DNA by path, or as data directly when registering.  Currently one can still use the DNA path when installing a happ, but this is deprecated in favor of allways using the hash.

This PR adds support for these calls.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@holochain/conductor-api",
-  "version": "0.0.1-dev.15",
+  "version": "0.0.1-dev.16",
   "description": "Encode/decode messages to/from the Holochain Conductor API over Websocket",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@holochain/conductor-api",
-  "version": "0.0.1-dev.16",
+  "version": "0.0.1-dev.15",
   "description": "Encode/decode messages to/from the Holochain Conductor API over Websocket",
   "repository": {
     "type": "git",
--- a/test/e2e/index.ts
+++ b/test/e2e/index.ts
@@ -44,12 +44,23 @@ test('admin smoke test', withConductor(ADMIN_PORT, async t => {
   await admin.attachAppInterface({ port: 0 })
   await admin.deactivateApp({ installed_app_id })
 
-  const dnas = await admin.listDnas()
+  let dnas = await admin.listDnas()
   t.equal(dnas.length, 1)
 
   const activeApps3 = await admin.listActiveApps()
   t.equal(activeApps3.length, 0)
   // NB: missing dumpState because it requires a valid cell_id
+
+  // install from hash
+  const newHash = await admin.registerDna({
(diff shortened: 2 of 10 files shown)


zo-el (commented):
We should get this in **only after** updating the `list_dnas` endpoint to return DNA hash with `app_id` and `dna_nick`. Else it is inconvenient to actually use this update

zippy:
> We should get this in **only after** updating the `list_dnas` endpoint to return DNA hash with `app_id` and `dna_nick`. Else it is inconvenient to actually use this update

Why?  You don't actually have to use this, it's completely backward compatible.  Not that we shouldn't get that in fast too!  But I want to get the ability to register DNA's from the DnaFile in.

zo-el:
> > We should get this in **only after** updating the `list_dnas` endpoint to return DNA hash with `app_id` and `dna_nick`. Else it is inconvenient to actually use this update
> 
> Why? You don't actually have to use this, it's completely backward compatible. Not that we shouldn't get that in fast too! But I want to get the ability to register DNA's from the DnaFile in.

The install_hosted_happ is blocked by the ability to recognize the servicelogger DNA from the list of DNA's. And the only use-case for this feature would be for installing servicelogger right now. So I think we should get both in together so that I can be tested out soon.

test/e2e/index.ts
```diff
@@ -34,15 +43,68 @@ test('admin smoke test', withConductor(ADMIN_PORT, async t => {
...
+
+  let dnas = await admin.listDnas()
+  t.equal(dnas.length, 1)
 
   const activeApps3 = await admin.listActiveApps()
   t.equal(activeApps3.length, 0)
   // NB: missing dumpState because it requires a valid cell_id
+
+  // install from hash
+  const newHash = await admin.registerDna({
```
zo-el, line 55: Quick Question: 
Just to confirm.
If I use this `newHash` and run another `admin.registerDna({source: {newHash}, properties: {uuid: 111})`
Will this new DNA that gets installed have the old properties of the newHash DNA? or is the source code the same as the first hash installed?

zo-el (approved):
Looks Good. Just left one clarification question.

test/e2e/index.ts
zippy: It should chain.  I.e. it should have the properties of the parent, not the root.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@holochain/conductor-api",
-  "version": "0.0.1-dev.15",
+  "version": "0.0.1-dev.16",
```
maackle:
```suggestion
  "version": "0.0.1-dev.15",
```

Unless you're doing things differently, I don't think you want to bump the version in the PR, `npm version` will do this automatically for you.

Also, it seems like next release could be a full patch release. I think now that we've hit the big-time we're ready to officially release `0.0.1` :smile:

maackle (approved):
Looks good

PR #45, opened 2021-01-19
Add API reference documentation

mjbrisebois:
Adds markdown files documenting the API reference for `AdminWebsocket`, `AppWebsocket`, and `WsClient`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ npm install --save-exact @holochain/conductor-api
   }, 30000)
 ```
 
+## API Reference
+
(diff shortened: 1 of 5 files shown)


maackle:
This looks great, but how would these docs be consumed by the typical user? Would they get built and published somewhere?

Is there a way to get these closer to the code, so they'll be more likely to stay up to date?

zippy:
This would be good question to answer.  It's an important second step, but for now lets at least get this merged.

zippy (approved):

PR #47, opened 2021-02-15
fix siglanCb logic

peeech:


zippy (approved):

timotree3 (approved):

Connoropolous:
just for reference, this is the same as #42 so close that one

src/websocket/client.ts
```diff
@@ -76,18 +76,22 @@ export class WsClient {
...
+              const decodedPayload = signalTransform(decodedMessage.App[1]);
 
-            // Return a uniform format to UI (ie: { type, data } - the same format as with callZome and appInfo...)
-            const signal: AppSignal = { type: msg.type , data: { cellId: decodedCellId, payload: decodedPayload }};
-            signalCb(signal);
+              // Return a uniform format to UI (ie: { type, data } - the same format as with callZome and appInfo...)
+              const signal: AppSignal = { type: msg.type , data: { cellId: decodedCellId, payload: decodedPayload }};
+              signalCb(signal);
+            } else {
+              console.error(`Received signal but no signal callback was set in constructor`);
```
Connoropolous: should this be a console.error? not setting a signalCb seems like a common use possibility

peeech: the way I see it is:
- if holochain is sending signals and there's no signalCB then it is an error
So not setting a signalCb in an application that does not use signals is fine and will never log an error

thoughts?

Connoropolous: Mostly fair point. One more thing to consider is that for a given holochain conductor, you wish (and I have done this) to construct multiple clients at different points and places in the code. Some may attach the callback listener, and provide the one and only handler, and others may not. The one that doesn't will log this error despite you having another one which is handling them

peeech: Yes, I see. So maybe a better solution would be to `console.log` it once during existence of this class instance? Easy to do and quite elegant I think.

timotree3: That sounds great!

Connoropolous: Yep!

peeech: cf5cb68 should do the job

src/websocket/client.ts
```diff
@@ -12,10 +12,12 @@ import { AppSignal, AppSignalCb, SignalResponseGeneric } from '../api/app'
 export class WsClient {
   socket: Websocket
   pendingRequests: Record<string, { fulfill: Function }>
+  noSignalCbFlag: Boolean
```
timotree3: Suggesting a longer variable name and changing the type from `Boolean` (a class) to `boolean` (typescript primitive)
```suggestion
  alreadyWarnedNoSignalCb: boolean
```

timotree3 (approved):
Newest commit looks good and I have a couple of suggestions should you choose to accept them

timotree3:
You're gonna need to update the usages of `noSignalCbFlag` to the new name as well

src/websocket/client.ts
```diff
@@ -76,18 +78,24 @@ export class WsClient {
...
+              const decodedCellId = decodedMessage.App[0];
+              // Note:In order to return readible content to the UI, the signal payload must also be decoded.
+              const decodedPayload = signalTransform(decodedMessage.App[1]);
+
+              // Return a uniform format to UI (ie: { type, data } - the same format as with callZome and appInfo...)
+              const signal: AppSignal = { type: msg.type , data: { cellId: decodedCellId, payload: decodedPayload }};
+              signalCb(signal);
+            } else {
+              if (!hw.noSignalCbFlag) console.log(`Received signal but no signal callback was set in constructor`);
+              hw.noSignalCbFlag = true;
```
timotree3: Two more variable usages to update
```suggestion
              if (!hw.alreadyWarnedNoSignalCb) console.log(`Received signal but no signal callback was set in constructor`);
              hw.alreadyWarnedNoSignalCb = true;
```

Connoropolous (approved):

PR #52, opened 2021-02-24
Update to latest holochain with Bundles

zippy:
Adds in all the types and functions needed to use the new bundle format in latest Holochain conductor API .


Connoropolous (commented):
I don't have a lot of bandwidth to review this, but my comment would be that with changes of this sort happening, it'd be really useful to have a little "glossary" or something that describes the new terms like...
"bundle"
"slot"
etc. 
and just a short summary of how to build a mental model of these things and use them.

zippy:
> I don't have a lot of bandwidth to review this, but my comment would be that with changes of this sort happening, it'd be really useful to have a little "glossary" or something that describes the new terms like...
> "bundle"
> "slot"
> etc.
> and just a short summary of how to build a mental model of these things and use them.

Yes, @Connoropolous  I will definitely upgrade the PR.

freesig (approved):

PR #54, opened 2021-03-08
Fix types

guillemcordoba:


src/api/admin.ts
```diff
@@ -214,8 +220,8 @@ type DnaManifest = {
 }
 
 export type DnaBundle = {
-    manifest: DnaManifest,
-    resources: ResourceMap,
+    manifest: DnaManifest;
```
zippy, line 223: Is this really supposed to be a ;?

guillemcordoba: Yeap! Actually autoformatting with prettier fixes a lot of format issues that exist in this file at the moment. I could commit those if you wanted.

zippy: That would be great because then it's standardized.  Thanks!

zippy (approved):
Thanks for catching all these mistakes!

PR #57, opened 2021-03-17
Don't mutate zome call args

timotree3:
This fixes an issue that was showing up with Holo Hosted Apps. In Holo's infrastructure we're retrying zome calls when one fails under certain conditions, but the second one was always failing due to this bug. See the added test for an example


maackle (approved):
Oh, Javascript, bless your heart

PR #64, opened 2021-04-05
Release v0.0.3: add uid to InstallAppBundleRequest

zo-el:
This is an update for the updates in holochain completed in the following PR:
- https://github.com/holochain/holochain/pull/726
- https://github.com/holochain/holochain/pull/727

### DIFF
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -17,7 +17,7 @@ export type GenerateAgentPubKeyRequest = void
 export type GenerateAgentPubKeyResponse = AgentPubKey
 
 export type RegisterDnaRequest = {
-  uuid?: string,
+  uid?: string,
   properties?: DnaProperties,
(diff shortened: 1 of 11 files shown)


src/api/admin.ts
```diff
@@ -130,6 +131,9 @@ export type InstallAppBundleRequest = {
     /// Include proof-of-membrane-membership data for cells that require it,
     /// keyed by the CellNick specified in the app bundle manifest.
     membrane_proofs: {[key: string]: MembraneProof},
+
+    /// Optional fixed UID. May be overridden during installation.
```
zippy:
```suggestion
    /// Optional global UID override.  If set will override the UID value for all DNAs in the bundle.
```

zippy (approved):
Just one updated comment suggestion

PR #69, opened 2021-05-12
Removed `Buffer.from`, fixed types to align with msgpack API

guillemcordoba:
All tests passing:

```
    ✔  should be truthy
    ✔  should be truthy
    ✔  should be strictly equal
    ✔  should be strictly equal
        Conductor exited with code null
        Conductor exited with code null
    ✔  should be strictly equal

passed: 55  failed: 0  of 55 tests  (25.5s)

All of 55 tests passed!
```

### DIFF
--- a/src/websocket/app.ts
+++ b/src/websocket/app.ts
@@ -49,7 +49,7 @@ export class AppWebsocket implements AppApi {
     = this._requester('zome_call_invocation', callZomeTransform)
 }
 
-const callZomeTransform: Transformer<CallZomeRequestGeneric<any>, CallZomeRequestGeneric<Uint8Array>, CallZomeResponseGeneric<Buffer>, CallZomeResponseGeneric<any>> = {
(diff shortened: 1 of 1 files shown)


src/websocket/app.ts
```diff
@@ -49,14 +49,14 @@ export class AppWebsocket implements AppApi {
     = this._requester('zome_call_invocation', callZomeTransform)
 }
 
-const callZomeTransform: Transformer<CallZomeRequestGeneric<any>, CallZomeRequestGeneric<Buffer>, CallZomeResponseGeneric<Buffer>, CallZomeResponseGeneric<any>> = {
-  input: (req:  CallZomeRequestGeneric<any>): CallZomeRequestGeneric<Buffer> => {
+const callZomeTransform: Transformer<CallZomeRequestGeneric<any>, CallZomeRequestGeneric<Uint8Array>, CallZomeResponseGeneric<Buffer>, CallZomeResponseGeneric<any>> = {
```
zippy: The `CallZomeResponseGeneric` still has a `Buffer` in it.  Not sure that's right.

guillemcordoba: Yeap thanks, fixed it. Actually I think the whole library would benefit from removing Buffer as one of its core main types, as this is mainly going to be used in UIs.

timotree3 (approved):
Thanks!

PR #72, opened 2021-06-16
List all apps

guillemcordoba:
Depends on https://github.com/holochain/holochain/pull/832


src/api/admin.ts
```diff
@@ -150,6 +150,9 @@ export type ListCellIdsResponse = Array<CellId>
 export type ListActiveAppsRequest = void
 export type ListActiveAppsResponse = Array<InstalledAppId>
 
+export type ListInActiveAppsRequest = void
```
zippy:
```suggestion
export type ListInactiveAppsRequest = void
```

src/api/admin.ts
```diff
@@ -150,6 +150,9 @@ export type ListCellIdsResponse = Array<CellId>
...
+export type ListInActiveAppsResponse = Array<InstalledAppId>
```
zippy:
```suggestion
export type ListInactiveAppsResponse = Array<InstalledAppId>
```

src/websocket/admin.ts
```diff
@@ -68,6 +68,8 @@ export class AdminWebsocket implements Api.AdminApi {
     = this._requester('list_cell_ids')
   listActiveApps: Requester<Api.ListActiveAppsRequest, Api.ListActiveAppsResponse>
     = this._requester('list_active_apps')
+  listInactiveApps: Requester<Api.ListInActiveAppsRequest, Api.ListInActiveAppsResponse>
```
zippy:
```suggestion
  listInactiveApps: Requester<Api.ListInactiveAppsRequest, Api.ListInactiveAppsResponse>
```

zippy (approved):

guillemcordoba:
@zippy I don't like the way the API looks with the weird `admin.listApps({status_filter: {'Active': null}})`, but looks like it's an upstream problem of messagepack-serde (see https://github.com/holochain/holochain/pull/279 and https://github.com/3Hren/msgpack-rust/issues/250). What should be done here? For now everything works, but with the weird API. 

Alternatively, we could accept a generic string in the filter and return an error if it doesn't match 'active' or 'inactive'. Yes I think that's better, let me change to that.

src/api/admin.ts
```diff
@@ -150,7 +150,7 @@ export type ListCellIdsResponse = Array<CellId>
 export type ListActiveAppsRequest = void
 export type ListActiveAppsResponse = Array<InstalledAppId>
 
-export type AppStatusFilter =  {'Active': null }| {'Inactive': null};
+export type AppStatusFilter =  'active' | 'inactive';
```
Connoropolous: Typescript will support nice code preview with this syntax right? No need to make it an enum structure?

Connoropolous (commented):
I like that you're hiding the awkward series serialization. I still can't believe they don't have reasonable serialization for flat enums

src/api/admin.ts
guillemcordoba: Oh yeah it will give good hints, but that's a good point also. I like the enum better.

Issue #21, opened 2020-10-19
setup code formatter so that PRs don't bounce around personal settings/preferences


Connoropolous:
and harm PR readability

jost-s:
I've submitted a suggestion.

jost-s:
Code formatter has been merged, so I'm closing this. Please re-open if it keeps causing problems, @Connoropolous.

PR #79, opened 2021-08-04
Add eslint as code formatter

jost-s:
Regarding #21 

I've added eslint and linted the code based on predominant style (single and double quotes as well as semi-colons and no semi-colons were mixed). We can swap that around depending on your preference =)

We can add prettier as well if you want.

I replaced a few unspecific types.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+  'env': {
+    'node': true,
+  },
+  'extends': [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended'
+  ],
+  'parser': '@typescript-eslint/parser',
+  'parserOptions': {
+    'ecmaVersion': 12,
+    'sourceType': 'module',
+  },
+  'plugins': ['@typescript-eslint'],
+  'rules': {
+    'comma-dangle': ['error', { 'objects': 'always-multiline', 'functions': 'never', 'arrays': 'never', 'imports': 'never' }],
+    'indent': ['error', 2],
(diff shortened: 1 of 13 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,24 @@
...
+  'parser': '@typescript-eslint/parser',
+  'parserOptions': {
+    'ecmaVersion': 12,
+    'sourceType': 'module',
+  },
+  'plugins': ['@typescript-eslint'],
+  'rules': {
+    'indent': ['error', 2],
+    'linebreak-style': ['error', 'unix'],
+    'quotes': ['error', 'single'],
```
zippy: I would rather not enforce one quote style.  There are often reasons to use one vs the other, i.e. depending on if you want to quote or not quote the singe or double quotes in a particular string.

zippy (changes requested):
This is great, thank you.  Just one request, that we don't force a quote style.   I'm fine with having them cleaned up in this one PR but in general I like the flexibility so that I can do both strings like this:
```
x = "test = 'fish'"
y = '<tag attr="dog">'
```
depending on what I don't want to have to quote.

jost-s:
Hi @zippy, in those cases where different quotes are needed, the back tick can be used
```javascript
x = `test = 'fish'`
y = `<tag attr="dog">`
```
Nevertheless I've removed the quote rule and reverted the affected lines.

Ready for review again.

zippy (approved):
Looks great, just the packge.lock merge conflicts and ready to go.

jost-s:
@zippy Super, resolved. A few more files were linted after the most recent commit.

PR #82, opened 2021-09-21
Introduce Launcher Environment

guillemcordoba:
This PR is targeting the next release for the launcher. In a normal development environment, this PR won't affect the behavior of this library, so this is really a non-breaking change.

When this library is imported, it will automatically perform a `GET` call to `/.launcher-env.json`, which is the file that the launcher sets. If it finds it, it will override the ports for both AppWebsocket and AdminWebsocket, also and installded_app_info when request appInfo. This assumes that in the future you should only be able the information about the application in which the UI is installed.

The result is that everything automagically works for the consuming developer, and they get a warning about it when they are developing so they can know a bit better what to expect.


zippy:
merged into hc-106

PR #83, opened 2021-09-21
Updates to hc 107 and add Launcher compatibility

zippy:
As well as updating to hc-107, this PR is targeting the next release for the launcher. In a normal development environment, this PR won't affect the behavior of this library, so this is really a non-breaking change.

When this library is imported, it will automatically perform a GET call to /.launcher-env.json, which is the file that the launcher sets. If it finds it, it will override the ports for both AppWebsocket and AdminWebsocket, also and installed_app_info when request appInfo. This assumes that in the future you should only be able to read the information about the application in which the UI is installed.

The result is that everything automagically works for the consuming developer, and they get a warning about it when they are developing so they can know a bit better what to expect.


default.nix
```diff
@@ -1,21 +1,36 @@
+{
```
zo-el, line 1: The dev branch will keep updating. Why do we want to move away from using revs?

docs/API_adminwebsocket.md
```diff
@@ -176,6 +176,13 @@ Request a new App to be created using the given Agent and DNAs.
 
 Returns a `Promise` for the corresponding response.
 
+## `<AdminWebsocket>.installApp({ installed_app_id })`
```
zo-el: Did you mean to document `uninstallApp`?

test/e2e/fixture/zomes/foo/Cargo.toml
```diff
@@ -10,4 +10,4 @@ crate-type = [ "cdylib", "rlib" ]
 
 [dependencies]
 serde = "=1.0.123"
-hdk = {git ="https://github.com/holochain/holochain", rev = "c5dbdf28825927106bc32d186dd54f20d35df468", package = "hdk"}
+hdk = "0.0.106"
```
zo-el:
```suggestion
hdk = "=0.0.106"
```

zo-el: Else it will get overridden when we have a new release. Unless that's what we want

zo-el (changes requested):
Few small changes were requested. Else its good

zo-el (approved):
Looks good

PR #85, opened 2021-10-27
Prevent console.error from going off because it causes some problems in test code

timotree3:
For example, [this CI run](https://github.com/holochain/elemental-chat-ui/runs/4025567093?check_suite_focus=true#step:5:63) fails to due to some `getFileName` error which I think is caused by the usage of `console.error`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,5 +20,6 @@ module.exports = {
     'semi': ['error', 'never'],
     '@typescript-eslint/no-explicit-any': 'off',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/no-empty-function': 'off'
   },
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
       },
       "devDependencies": {
         "@detools/tap-diff": "^0.2.2",
-        "@types/node": "^14.0",
+        "@types/node": "^14.17.32",
         "@types/tape": "^4.13.0",
         "@types/ws": "^7.2.4",
         "@typescript-eslint/eslint-plugin": "^4.29.0",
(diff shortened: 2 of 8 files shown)


timotree3:
I may have merged #79 prematurely. When running tests on develop, I saw that 0 out of 0 tests passed, which counts as success :/ This PR now fixes that and additionally updates the lints to apply to the `tests/` folder

timotree3:
I actually think jest is pointing us to a valid criticism of holochain-conductor-api. I don't think libraries should log things, and they should definitely not log errors under expected and normal circumstances. If we're looking for the proper solution here, I propose that we propagate this error properly. Perhaps a good solution would be to capture this error but not print it, and then edit the error path of `AppWebsocket.connect` to detect this error, and include it. e.g.

> Exception: Could not connect to holochain app websocket at localhost:xxxx
> Note: This may be because we are in a launcher environment and failed to detect that. Reason: ${launcherDetectionError}

guillemcordoba:
So the problem is really that we can't hide the error in the browser, see https://stackoverflow.com/questions/12915582/eliminate-404-url-error-in-console and https://github.com/verlok/vanilla-lazyload/issues/460. When we do the `fetch` from a browser but not inside the launcher environment, we are going to get an error in the console, and we can't do anything for that. To avoid people being confused about that, I included the console.warn just after that to let the devs know that they don't have to worrry about that 404 error.

What we could do is do the fetch in the first `AppWebsocket.connect()`, and then it's not a side effect from importing the library... I like that more, but it still has the same problem with the 404 error, we can't really capture it.

Would you be okey if I did that change?

The thing you are running into is that Jest is not really a browser environment, so it shouldn't even do the fetch, but Jest tries to simulate a browser so we have thing like window set. Jest is weird...

timotree3:
I could imagine wanting to test launcher autodetection using jest. maybe the best approach would be to revert to before this PR but just put the `console.warn` and `console.error` within an if that we are not inside jest?

timotree3:
I will test out this version as requested.

guillemcordoba (approved):

PR #86, opened 2021-10-27
Use cachix for CI

timotree3:
We are doing this in many holochain org repos and it massively improves CI times


timotree3:
Looks like in this particular case the version of holochain we are using here is not cached on cachix, but I think this is still worth merging for the future.

guillemcordoba (approved):

PR #93, opened 2021-11-23
Crash when setting up tests should cause non-zero exit code

harlantwood:
Without this, if we crash while setting up tests, the test command will return a zero exit code, and builds will report as green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "doc": "typedoc",
     "prepare": "npm run build",
     "prepublishOnly": "npm test",
-    "test": "npm run test:raw | tap-diff",
+    "test": "set -o pipefail && npm run test:raw | tap-diff",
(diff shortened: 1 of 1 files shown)


harlantwood:
Inspired by a fix to where, which was broken here (but CI green): 
https://github.com/lightningrodlabs/where/actions/runs/1387171777

Then red once we merged a similar fix (https://github.com/lightningrodlabs/where/pull/22/files):
https://github.com/lightningrodlabs/where/actions/runs/1387248783

harlantwood:
@zippy

jost-s (commented):
In the new repo I got rid of `tap-diff` and all the other TAP reporters. The issue is resolved then, `tape` returns the correct exit code.

jost-s:
Closing this as this repo will be archived.

PR #94, opened 2021-12-06
Dump full state, importing lots of types from https://github.com/holochain-open-dev/core-types

guillemcordoba:
In the effort to make the Holochain Playground compatible with a real Holochain conductor, I've had to create all the types that the `dump_full_state` function is returning. I thought it would be useful to include them at the @holochain/conductor-api level, so that everyone gets them. Let me know if you'd prefer to keep them at https://github.com/holochain-open-dev/core-types.


Connoropolous:
@guillemcordoba this still under consideration for review?

guillemcordoba:
Yeap!

jost-s (approved):
This is brilliant, thanks so much @guillemcordoba !!!

I've included this work into the new repo of the JS client.

jost-s:
I'm closing this PR regardless, because we're archiving the repo in favor of the new one.

jost-s:
My mistake, we're not archiving this repo but only changing its name. Regardless, this branch's changes are included.

PR #101, opened 2022-01-27
Export appBundleWithProperties utility

timotree3:
Currently, there is no easy way to install apps and specify properties at the same time, without using `registerDna`. This PR adds a utility function that converts an `AppBundleSource` like `{ path: "./test.happ" }` to `{ bundle: { ... } }` by parsing the happ file at that path. It then edits the bundle to add some properties, since I see that as the primary use-case for this functionality. 

This function only works in a Node environment. It would be nice if as part of our test suite, we made sure things still compiled well in a browser environment. For now I'll use a local UI to test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,6 +289,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b03d17f364a3a042d5e5d46b053bbbf82c92c9430c592dd4c064dc6ee997125"
 
 [[package]]
+name = "linked-hash-map"
+version = "0.5.4"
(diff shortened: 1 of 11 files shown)


timotree3:
Just tested this with a happ UI. Encountered this error
```
This dependency was not found:

* fs/promises in ./node_modules/@holochain/client/lib.es/bundle.js
```

timotree3:
Nice! Tested with webpack and after the latest commit, it was able to compile despite the fact that it conditionally imports Node modules.

jost-s (approved):
Thanks for adding this function, @timotree3! Yeah, I'm planning to add browser tests.

PR #117, opened 2022-07-05
Updates nomenclature in types from Header/Element to Action/Record

zippy:


jost-s (approved):
Thanks, @zippy! LGTM

Ah, please restore the `package-lock.json`.

PR #120, opened 2022-07-07
HoloHash in CellId could be more specific

Connoropolous:


jost-s (approved):
thank you @Connoropolous

PR #121, opened 2022-07-08
Fix race condition for the installed app id in the launcher

guillemcordoba:
This is producing problems in the REA-playspace app.


jost-s (commented):
LGTM

PR #127, opened 2022-08-11
update docs for AdminWebsocket

mhuesch:
de/activateApp=>dis/enableApp

was deprecated back in `holochain` 0.0.102: https://github.com/holochain/holochain/blob/33f26d8f08e3c61f7981cd0a7441d479386da17a/crates/holochain/CHANGELOG.md#added

### DIFF
--- a/docs/API_adminwebsocket.md
+++ b/docs/API_adminwebsocket.md
@@ -22,6 +22,23 @@ Create a new connection to the given URL.
 Returns a `Promise` for a new connected instance of `AdminWebsocket`.
 
 
+## `<AdminWebsocket>.activateApp({ installed_app_id })`
+
+| :exclamation: This method is deprecated. Please use `enableApp` instead  |
+|--------------------------------------------------------------------------|
+
+Send a request to activate an installed app.
+
+- `installed_app_id` - a `string`
+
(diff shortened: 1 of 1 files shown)


jost-s:
Thanks for the update. The API has been deprecated indeed, but it still exists https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api/enum.AdminRequest.html#variant.ActivateApp

As long as it's there, I don't want to remove it from the client. We can keep this PR until the API is actually removed.

mhuesch:
got it! thx. what do you think about adding entries for the ena/disableApp methods, and adding deprecation warnings on de/activateApp?

I've just used the wrong ones a couple times b/c the right ones weren't documented, and there isn't a deprecation warning. wonder if others have encountered the same.

jost-s:
Absolutely, that would be great! Go ahead and add them if you like, or else I'll have time for that next week.

mhuesch:
@jost-s just pushed another commit making those changes. lmk your thoughts :pray: :raised_hands:

jost-s (approved):
@mhuesch Super, thanks a bunch!

mhuesch:
@jost-s cool, you're welcome! :raised_hands:

PR #133, opened 2022-09-26
Tauri windows

matthme:
Necessary changes that allow opening apps in tauri windows. The proposed changes should not interfere with the currently implemented functionality for opening apps in the browser but are complementary.

### DIFF
--- a/src/environments/launcher.ts
+++ b/src/environments/launcher.ts
@@ -15,7 +15,7 @@ async function fetchLauncherEnvironment(): Promise<
   LauncherEnvironment | undefined
 > {
   const env = await fetch(LAUNCHER_ENV_URL);
-  console.log("@holochain/client@fetchLauncherEnvironment: env: ", env);
+
 
   if (env.ok) {
     const launcherEnvironment = await env.json();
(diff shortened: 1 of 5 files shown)


matthme:
Uff :/. The test fails because ts-node complains about `window` not being defined (which is actually part of the functionality of determining whether or not the client is within a tauri context...). Any ideas how to fix this?

src/environments/launcher.ts
```diff
@@ -34,6 +34,7 @@ async function fetchLauncherEnvironment(): Promise<
   }
 }
 
+const isTauriWindow = !!(window as any).__TAURI__;
```
guillemcordoba: Here you can do it like this:
```ts
const isBrowser = typeof window !== "undefined";
const isTauriWindow = isBrowser && !!(window as any).__TAURI__;
```

guillemcordoba:
This doesn't do the invokes yet, so it's not ready right?

matthme:
This is only for allowing to open apps in tauri windows, independently of code signing. So that part is ready but if you prefer to make one PR including the zome call signing then we should wait yes.

guillemcordoba:
@jost-s ?

jost-s:
I was gonna add zome call signing to the JS client this week. Did you want to add signing with this PR, is that what you mean?

src/environments/launcher.ts
```diff
@@ -49,7 +50,7 @@ if (isBrowser && !isJest) {
 export async function getLauncherEnvironment(): Promise<
   LauncherEnvironment | undefined
 > {
-  if (isBrowser) {
+  if (isBrowser || isTauriWindow) {
```
Connoropolous, line 53: I might be reading this logic wrong, but it looks funny because isTauriWindow can never be true if isBrowser is not true. So it seems it would change nothing

matthme: Hahaha! This PR is becoming a farce. You are totally right of course. When I started with the code changes I assumed there will be a difference between browser mode and tauri mode but I ended up using a solution that does the same in both cases.

Connoropolous: hehehe

PR #139, opened 2022-11-29
Add AppAgentWebsocket class and `AppAgentClient` interface

robbiecarlton:
This adds an interface, `AppAgentClient`, along with a class that implements that interface `AppAgentWebsocket` as part of work to unify the holochain and holo client apis.

Happ devs will now be able to write the majority of their client code using the `AppAgentClient` interface, completely agnostic as to whether they're in a pure holochain or holo context.

For more details, see: 
https://hackmd.io/BRe8ohENSvyGTBglk7Va5A
and
https://hackmd.io/fRfmYnGPSyuHkeu_k77Nww


src/api/app_agent/websocket.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  constructor(
+    appWebsocket: AppWebsocket,
+    installedAppId: InstalledAppId
+  ) {
+    super()
+    this.appWebsocket = appWebsocket;
+    this.installedAppId = installedAppId
+
+    this.appWebsocket.on('signal', signal => this.emit('signal', signal))
```
robbiecarlton: Currently this forwards ALL signals, it doesn't filter based on `installed_app_id`. I didn't add that filter, as it would require an additional call to `appInfo` for every signal, but it's possible it's worth it.

Curious your thoughts on this.

src/api/app_agent/websocket.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  async appInfo (): Promise<AppInfoResponse> {
+    return this.appWebsocket.appInfo({
+      installed_app_id: this.installedAppId
+    })  
+  }
+
+  async callZome (request: AppAgentCallZomeRequest, timeout?: number): Promise<CallZomeResponse> {
+    if (request.role_id) {
+      const appInfo = await this.appInfo()
```
robbiecarlton: We could cache appInfo relatively easily (we only really need to update the cache when creating a new clone cell, even though the presence and status of the clones might change through admin calls, the mapping from role_id to clone won't break)
BUT, given that this "call by `role_id`" feature is going to be pushed down into the conductor api sooner or later, I decided it wasn't worth adding the extra complexity in the meantime here.

Let me know if you think it's worth it.

README.md
```diff
@@ -49,6 +49,30 @@ npm install --save-exact @holochain/client
...
+  const TIMEOUT = 12000
+  // default timeout is set to 12000
+  const appWs = await AppWebsocket.connect(`ws://localhost:${appPort}`, 12000, signalCb)
+
+  const client = new AppAgentWebsocket(appWs, 'installed_app_id')
+
+  // default timeout set here (30000) will overwrite the defaultTimeout(12000) set above
+  await client.callZome({
+   cap: null,
+   role_id: 'dnas_role_id', // role_id is unique per app, so you can unambiguously identify your dna with role_id in this client,
```
guillemcordoba, line 68: This is now called role name right?

robbiecarlton: Hmm, just saw that.

@jost-s - should I update `role_id` to `role_name` across this whole repo? Or leave that to you in a separate pr

jost-s: @robbiecarlton I'll either do this in signed zome calls or in a separate branch.

docs/API_appagentwebsocket.md
```diff
@@ -0,0 +1,95 @@
...
+# `new AppAgentWebsocket( appWs )`
+A class for interacting with Conductor's app API, restricted to a specific installed app.
+This is useful for simplifying zome and app info calls, and especially because it shares an interface (`AppAgentClient`) with the holo WebSdk client, meaning you can use this client to write the majority of your UI agnostic as to wether it's in a pure holochain or holo context.
+
+- `appWs` - an instance of AppWebsocket, connecting to an app port on your conductor
+
+**Instance properties**
+
+- `<AppAgentWebsocket>.appWebsocket` - the AppWebsocket instance that all calls are forwarded to.
+- `<AppAgentWebsocket>.installedAppId` - the installedAppId that all calls are restricted to
```
robbiecarlton:
```suggestion
- `<AppAgentWebsocket>.installedAppId` - the InstalledAppId that all calls are restricted to
```

jost-s (approved):
Fantastic, @robbiecarlton, thanks a bunch!

PR #144, opened 2022-12-08
Improved types for `AppAgentWebsocket`

guillemcordoba:
- `AppAgentWebsocket` doesn't extend `Emittery` anymore to prevent users from calling `emit('signal') themselves`.
- The `AppAgentClient` interface and `Emittery` instance now have fully type-safe events. Things like `appClient.on('some-event-that-doesnt-exist')` will throw a compile time error.


robbiecarlton (approved):
LGTM, thanks!

jost-s (approved):

PR #147, opened 2022-12-14
AppAgentWebsocket uses launcher env INSTALLED_APP_ID when present

robbiecarlton:
`AppAgentWebsocket` now checks the launcher env for `INSTALLED_APP_ID` and uses it when present. This means `AppAgentWebsocket` will work as expected in UIs running in launcher.

### DIFF
--- a/src/api/app-agent/websocket.ts
+++ b/src/api/app-agent/websocket.ts
@@ -50,14 +50,11 @@ export class AppAgentWebsocket implements AppAgentClient {
 
   emitter = new Emittery<AppAgentEvents>();
 
-  constructor(
-    appWebsocket: AppWebsocket,
-    installedAppId: InstalledAppId
-  ) {
+  constructor(appWebsocket: AppWebsocket, installedAppId: InstalledAppId) {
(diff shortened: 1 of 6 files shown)


src/api/app-agent/websocket.ts
```diff
@@ -49,15 +50,31 @@ export class AppAgentWebsocket implements AppAgentClient {
...
     this.appWebsocket = appWebsocket;
+
     this.installedAppId = installedAppId;
 
     this.appWebsocket.on("signal", (signal) =>
       this.emitter.emit("signal", signal)
     );
   }
 
+  static async connect(
```
robbiecarlton: I don't love calling this connect. Given that the normal usage is 
```
const appWs = AppWebsocket.connect()
const appAgentWs = AppAgentWebsocket.connect(appWs)
```
It's a little weird to say we're connecting twice, really we're connecting once.

What we're really doing is currying, but calling this function `.curry` is probably not great devX :)

Open to suggestions if someone has a better name.

guillemcordoba (approved):

PR #148, opened 2022-12-15
Tauri zome call signing

matthme:
Before it gets forgotten that I did the launcher part on another branch. This branch contains the tested code-signing `callZomeTransform` for the launcher case. It also sets "crypto: false" in package.json so that bundlers don't try to import the NodeJS built-in "crypto" module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target
 node_modules
 /lib
 .tsbuildinfo
+.hc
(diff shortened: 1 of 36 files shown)


matthme:
Replaced with a PR into [refactor/0.1.0-beta-rc.1](https://github.com/holochain/holochain-client-js/tree/refactor/0.1.0-beta-rc.1)

PR #150, opened 2022-12-17
Network info & Launcher zome call signing

matthme:
I'm gonna test this branch now in the launcher and if it does, I'll remove the Draft status.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,10 +44,10 @@ jobs:
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
           restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
 
-      - name: Run all tests
-        id: build-and-test
-        run: nix-shell --run './run-test.sh'
+      # - name: Run all tests
+      #   id: build-and-test
+      #   run: nix-shell --run './run-test.sh'
 
       - name: Setup tmate session if build and test run failed
-        if: steps.build-and-test.outcome == 'failure'
+        # if: steps.build-and-test.outcome == 'failure'
         uses: mxschmitt/action-tmate@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,4 +45,9 @@ jobs:
           restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
 
       - name: Run all tests
+        id: build-and-test
         run: nix-shell --run './run-test.sh'
+
+      - name: Setup tmate session if build and test run failed
+        if: steps.build-and-test.outcome == 'failure'
+        uses: mxschmitt/action-tmate@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,10 +44,10 @@ jobs:
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
           restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
 
-      # - name: Run all tests
-      #   id: build-and-test
-      #   run: nix-shell --run './run-test.sh'
+      - name: Run all tests
+        id: build-and-test
+        run: nix-shell --run './run-test.sh'
 
       - name: Setup tmate session if build and test run failed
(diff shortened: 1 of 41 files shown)


matthme:
I must have made a mistake, it's still supposed to be a draft. I'll comment here once it's tested.

jost-s (approved):

matthme:
Resolved conflicts. I require `CapSecret | null` in `CallZomeRequestUnsigned` for signing in the Launcher (line 153 in [src/api/app/websocket.ts](https://github.com/holochain/holochain-client-js/pull/150/files#diff-1675df576bed96b69c8742fcd29a87eb3cbb14aab4dd4f5686a00fd449f5d357)). Please check that this does not lead to type conflicts in the non-launcher case @jost-s

matthme:
Also switching to moduleResolution "Node" because I get the same error with "Node16" as I got with "NodeNext"

jost-s:
> Resolved conflicts. I require `CapSecret | null` in `CallZomeRequestUnsigned` for signing in the Launcher (line 153 in [src/api/app/websocket.ts](https://github.com/holochain/holochain-client-js/pull/150/files#diff-1675df576bed96b69c8742fcd29a87eb3cbb14aab4dd4f5686a00fd449f5d357)). Please check that this does not lead to type conflicts in the non-launcher case @jost-s

No, it doesn't cause a problem. Does the launcher not return a cap secret for the call?

matthme:
> No, it doesn't cause a problem. Does the launcher not return a cap secret for the call?

You're right. I don't need it anymore, provided that zome calls made in a launcher environment always have cap_secret null. I didn't check that properly, it can actually be removed again.

jost-s:
> You're right. I don't need it anymore, provided that zome calls made in a launcher environment always have cap_secret null. I didn't check that properly, it can actually be removed again.

I'll remove it with the next patch.

jost-s:
Would it be
```ts
interface CallZomeRequestSignedTauri // Tauri requires a number array instead of a Uint8Array
  extends Omit<CallZomeRequestSigned, "cell_id" | "provenance" | "nonce"> {
  cell_id: [TauriByteArray, TauriByteArray];
  provenance: TauriByteArray;
  nonce: TauriByteArray;
  expires_at: number;
}
```
and
```ts
const signedZomeCall: CallZomeRequestSigned = {
  provenance: Uint8Array.from(signedZomeCallTauri.provenance),
  cap_secret: Uint8Array.from(signedZomeCallTauri.cap_secret),
  cell_id: [
    Uint8Array.from(signedZomeCallTauri.cell_id[0]),
    Uint8Array.from(signedZomeCallTauri.cell_id[1]),
  ],
  zome_name: signedZomeCallTauri.zome_name,
  fn_name: signedZomeCallTauri.fn_name,
  payload: Uint8Array.from(signedZomeCallTauri.payload),
  signature: Uint8Array.from(signedZomeCallTauri.signature),
  expires_at: signedZomeCallTauri.expires_at,
  nonce: Uint8Array.from(signedZomeCallTauri.nonce),
};
```
?

matthme:
No, this part can all stay as is. It's just the `| null` option in the `CallZomeRequestUnsigned` interface that I don't need anymore (pinged you on the relevant line).

jost-s:
Right, but if I remove `| null` from that type, `cap_secret` has to be set in the launcher arm of the if statement.
Please submit a small PR =)

matthme:
It already is: https://github.com/holochain/holochain-client-js/blob/77b537d7450e991a3e29614eae32fd97cd50b524/src/api/app/websocket.ts#L204

jost-s:
I mean set to a value, a Uint8Array. Having `CallZomeRequestSigned` with an optional `cap_secret` is misleading if devs want to do their own signing.

matthme:
Oh now I get it, because `CallZomeRequestSigned` extends `CallZomeRequestUnsigned`. Yes but I think then we have to leave it as is because I need to return cap_secret `null` I think. I cannot just come up with a random cap secret if there is none, right?

jost-s:
No, you can't. But I see now, JS calls always have to have a cap secret and Launcher ones will never have one. That's fine as it is then.

matthme:
What I think I forgot in this PR is the possibility that an app sends an already signed zome call in the launcher, so instead of
```
    if (isLauncher) {
```
I think it should be
```
    if (isLauncher && !("signature" in req)) {
```

jost-s:
Okay, I'll weave that in.

jost-s:
here we are
https://github.com/holochain/holochain-client-js/commit/0fd4c7a9d6508507c5908f2652fba6eb120e2dbf

PR #152, opened 2022-12-19
Omit provenance field

guillemcordoba:


jost-s:
@guillemcordoba How about a simple field `AppAgentWebsocket.agentPubKey` that becomes part of the constructor?

guillemcordoba:
@jost Hum but then how is an app supposed discover the agentpubkey and pass that in? By calling `appWebsocket.appInfo()`?

And also, when agentpubkey is included in the result of `AppInfo`, doesn't this constructor argument become a foot gun?

README.md
```diff
@@ -104,6 +103,7 @@ You need `holochain` and `hc` on your path, best to get them from nix with `nix-
 To perform the pre-requisite DNA compilation steps, and run the Nodejs test, run:
 ```bash
 nix-shell
+npm run build
```
jost-s: Why is building required before running the tests?

src/types.ts
```diff
@@ -9,6 +9,16 @@ export type AnyDhtHash = HoloHash;
 export type KitsuneAgent = Uint8Array;
 export type KitsuneSpace = Uint8Array;
 
+/** Base64 hash types */
+
+export type HoloHashB64 = string;
```
jost-s, line 14: This one is in `src/api/admin/types.ts` too. Please remove it there.

src/utils/base64.ts
```diff
@@ -0,0 +1,10 @@
+import { Base64 } from "js-base64";
+import { HoloHash, HoloHashB64 } from "../types.js";
+
+export function deserializeHash(hash: HoloHashB64): HoloHash {
```
jost-s: Can we rename these to `encodeHashTo64` and `decodeHashFromB64` or something similar? It has nothing to do with serialization.

README.md
guillemcordoba: Oh sorry it isn't? My bad, will remove

src/types.ts
guillemcordoba: Hum I think I removed it? I don't see it in the code...

jost-s: Sorry, I looked at the wrong branch.

jost-s (approved):

PR #153, opened 2022-12-20
fix `functions` parameter type in `authorizeNewSigningKeyPair`

weswalla:
There is an error with the `functions` parameter type in `authorizeNewSigningKeyPair`. It should be `[string, string][]` (an array of a `[string, string]` tuple) instead of `[[string, string]]` (a tuple inside a tuple), otherwise it leads to errors like this:
image


jost-s (approved):

jost-s:
I just came across this too =)

PR #155, opened 2022-12-21
Make myPubKey sync

guillemcordoba:


jost-s:
I'd make the constructors private, actually of all websockets. That prevents people from calling them with `new`. Are you fine with that?

jost-s (approved):
LGTM

PR #158, opened 2022-12-29
Optional LauncherEnv parameters

matthme:
The launcher admin window needs a way to use the js-client where the js-client recognizes that it's in a launcher environment, i.e. passes zome calls for signing to tauri, but at the same time does not take the websocket URL from `window.__HC_LAUNCH_ENV__` but uses the one passed as an argument to the `.connect()` method. This is because the launcher admin window needs to be able to connect to multiple holochain versions from within the same window.

My first solution used an optional `ignoreLauncherEnv` argument in the connect method (ddd457e01d6051de5d7edf2ffd2074d97d961ce4) but then I realized that it's probably cleaner to solve it by making the attributes of the `LauncherEnv` interface optional (250004179a0b0edaaab36da079be0f7a400fcd94).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 ### Removed
 ### Changed
+- All Launcher environment properties to be optional.
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


jost-s:
including this in branch fix/github-issues

PR #166, opened 2023-01-18
Refactored signal to align with rc.3

guillemcordoba:


src/api/client.ts
```diff
@@ -65,15 +65,16 @@ export class WsClient extends Emittery {
           // We have received a system signal, do nothing
           return;
         }
+        const encodedAppSignal = derializedSignal["App"];
```
Connoropolous: You could use SignalType.App here

src/api/app/types.ts
```diff
@@ -101,11 +101,17 @@ export const SignalType = {
...
 export type Signal = {
-  [SignalType.App]: [CellId, any];
+  [SignalType.App]: EncodedAppSignal;
   [SignalType.System]: unknown;
 };
+export type EncodedAppSignal = {
+  cell_id: CellId;
+  zome_name: string;
+  signal: Uint8Array;
+};
```
Connoropolous, line 167: What about introducing PermaLinks comments through to the Rust side code? Bumping into quite a few serialization issues ongoingly

jost-s: I'm not familiar with PermaLinks, what is that?

Connoropolous: oops, just meant like a github permalink... like this. 
Screenshot 2023-01-18 at 9 08 02 PM

PR #171, opened 2023-01-23
feat(nix): exclude launcher from holonix

steveej:


jost-s:
This workaround isn't needed any longer.

PR #173, opened 2023-02-01
stringify error responses such that they are readable in the browser console

matthme:
I kept getting errors of the kind `unresolved Promise rejection: [Object Object]` for failing zome calls which were really unhelpful and I figured out that stringifying the result in the error case solves that.


jost-s:
Are you okay with the slightly increased effort of try/catching?

matthme:
Yes I think I can live with it. You can close the PR if you like.

PR #183, opened 2023-04-13
feat(storage-info): Add storage info endpoint to the admin API client

ThetaSinner:


ThetaSinner:
The test isn't happy because `storage_info` doesn't exist in the `holochain` currently pointed to by holonix - still working out what to do about that

jost-s:
Until we have a holonix version for 0.2 we can only update the flake to reference `develop`. I don't want to do that.
I'll run the tests locally too and then we should be good. It's not a potential source of a lot of bugs =)

src/api/admin/types.ts
```diff
@@ -706,6 +706,43 @@ export interface FullStateDump {
...
+  cache_data_size: number;
+  cache_data_size_on_disk: number;
+  used_by: Array<InstalledAppId>;
+}
+
+/**
+ * @public
+ */
+export interface DnaStorageBlob {
+  Dna: DnaStorageInfo;
```
jost-s, line 726:
```suggestion
  dna: DnaStorageInfo;
```

jost-s (approved):
Great!

Tested locally with Holochain v0.2.0-beta-rc.4 with success.

It's quite late where you are, so I lower-case the property and merge and release this.

src/api/admin/types.ts
jost-s: This has to be fixed upstream in Holochain. The Conductor API should produce all lower case snake case fields/properties.

jost-s: fixed by https://github.com/holochain/holochain/pull/2234

jost-s:
> Until we have a holonix version for 0.2 we can only update the flake to reference `develop`. I don't want to do that. I'll run the tests locally too and then we should be good. It's not a potential source of a lot of bugs =)

Actually I pointed the flake to rc.4 after all.

PR #186, opened 2023-04-26
updated NetworkInfoRequest type

matthme:
Updates the `NetworkInfoRequest` type to the recent changes in the conductor API.


jost-s (approved):
Great, thanks! Haaa, I forgot to write a test for this call and therefore didn't realize the request parameters were wrong.

PR #192, opened 2023-06-08
fix: error handling for localhost

zeeshan595:
Passing `localhost` does not work for hc sandbox running inside nix develop. Use `127.0.0.1` instead. Not sure if this change is needed maybe we can just update the docs to capture that

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -111,6 +111,9 @@ export class WsClient extends Emittery {
    */
   static connect(url: string) {
     return new Promise<WsClient>((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


jost-s:
Yes, I've come across that too and changed all the tests to use 127.0.0.1. It's to do with [Node.js having changed the preferred DNS resolution from IPv4 to IPv6](https://github.com/nodejs/node/pull/39987). `localhost` will therefore resolve to `::1`, and if your happ isn't also using IPv6 which I suppose it isn't doing, then the address cannot be found.

Issue #195, opened 2023-07-24
The strange case of the wandering `this`


ThetaSinner:
I ran into an issue with a scaffolded hApp using Vue. The front-end after scaffolding doesn't actually have signals working so it's not super obvious that there's a problem. But when you try to use them you get an error logged to the console

```
Uncaught TypeError: Cannot read property 'has' of undefined
```

Following this error lead me to Emittery and an existing issue about this https://github.com/sindresorhus/emittery/issues/86. I'm 100% sure this should be fixed in Emittery but in the meantime you actually can't use signals in a Vue app without a hard-to-guess workaround that @mattyg has added to that issue.

The problem is with `this` not being bound to the Emittery class properly. The Emittery class then uses itself (`this`) as a key into a hash map. When a client class from this library that extends Emittery is put inside a proxy that all falls apart. It's worth noting that while you could keep this class outside a proxy, scaffolded hApps  use `provide`/`inject`  which is a sensible choice and automatically uses proxies so we're already pretty committed on that front.

We can fix this in our constructors by doing `this.on = this.on.bind(this);`. I've tested this in the constructor for `AppAgentWebsocket`.

I'd like to keep this issue open to track the upstream issue until it gets fixed, but for now we can fix this in the Holochain client and remove a really nasty gotcha for Vue users.

jost-s:
workaround is fine

PR #196, opened 2023-07-26
Fix proxy this for emittery

ThetaSinner:
This addresses https://github.com/holochain/holochain-client-js/issues/195

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -23,9 +23,9 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install nix
-        uses: cachix/install-nix-action@v19
+        uses: cachix/install-nix-action@v22
         with:
-          install_url: https://releases.nixos.org/nix/nix-2.13.3/install
+          install_url: https://releases.nixos.org/nix/nix-2.17.0/install
 
       - name: Set up cachix
         uses: cachix/cachix-action@v12
--- a/src/api/app-agent/websocket.ts
+++ b/src/api/app-agent/websocket.ts
@@ -45,6 +45,15 @@ export class AppAgentWebsocket implements AppAgentClient {
     this.appWebsocket = appWebsocket;
     this.emitter = new Emittery<AppAgentEvents>();
 
+    // Ensure all super methods are bound to this instance because Emittery relies on `this` being the instance.
+    // Please retain until the upstream is fixed https://github.com/sindresorhus/emittery/issues/86.
+    Object.getOwnPropertyNames(Emittery.prototype).forEach((name) => {
+        const to_bind = (this.emitter as unknown as {[key: string]: unknown})[name];
+        if (typeof to_bind === 'function') {
+          (this.emitter as unknown as {[key: string]: unknown})[name] = to_bind.bind(this.emitter);
+        }
+    });
+
     const env = getLauncherEnvironment();
     this.installedAppId = env?.INSTALLED_APP_ID || installedAppId;
 
--- a/src/api/app/websocket.ts
+++ b/src/api/app/websocket.ts
@@ -59,6 +59,15 @@ export class AppWebsocket extends Emittery implements AppApi {
     overrideInstalledAppId?: InstalledAppId
   ) {
     super();
+    // Ensure all super methods are bound to this instance because Emittery relies on `this` being the instance.
+    // Please retain until the upstream is fixed https://github.com/sindresorhus/emittery/issues/86.
+    Object.getOwnPropertyNames(Emittery.prototype).forEach((name) => {
+      const to_bind = (this as unknown as {[key: string]: unknown})[name];
+      if (typeof to_bind === 'function') {
+        (this as unknown as {[key: string]: unknown})[name] = to_bind.bind(this);
+      }
+    });
+
     this.client = client;
     this.defaultTimeout =
       defaultTimeout === undefined ? DEFAULT_TIMEOUT : defaultTimeout;
--- a/test/e2e/app-agent-websocket.ts
+++ b/test/e2e/app-agent-websocket.ts
@@ -124,7 +124,7 @@ test(
 );
 
 test(
-  "can receive a signal when proxied",
+  "can receive a signal when emitter is proxied",
   withConductor(ADMIN_PORT, async (t: Test) => {
     let resolveSignalPromise: (value?: unknown) => void | undefined;
     const signalReceivedPromise = new Promise(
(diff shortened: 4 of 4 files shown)


ThetaSinner:
I've struggled to reproduce this in Node tests. I can do it with JavaScript but whatever is happening there, TypeScript seems to fix. Without digging deeper into the Vue.js source I'm not sure what they are doing differently. Maybe somebody with better TypeScript knowledge can figure this one out.

I have installed this PR into my own project and it does fix the problem.

ThetaSinner:
Not sure what's happening with the macos build, is this a new problem that's worth me digging into or something known that we can address?

jost-s:
> Not sure what's happening with the macos build, is this a new problem that's worth me digging into or something known that we can address?

That's a problem with the nix install action. I upgraded it, that should fix it.

ThetaSinner:
Awesome, thank you!

jost-s:
@ThetaSinner Why do you want an AppAgentWs to be reactive? I see that @mattyg did that in mewsfeed, but I cannot see why that's necessary there, either. It seems like the AppAgentWs can just be a variable with `let` and doesn't need to be reactive.

mattyg:
> @ThetaSinner Why do you want an AppAgentWs to be reactive? I see that @mattyg did that in mewsfeed, but I cannot see why that's necessary there, either. It seems like the AppAgentWs can just be a variable with `let` and doesn't need to be reactive.

It needs to be reactive so I can initialize it in the vue app's root component. Then I can display content while waiting for it to initialize, and when its ready, reactively make it available to sub-components via provide / inject. So it needs to reactively update the client variable from  `undefined` to the actual AppAgentWs. (also this is how the scaffolder implements it)

I just ran into this issue in another context where I am passing the AppAgentWebsocket into new class as a constructor param, and then trying to listen to signals within that class. Unfortunately this PR doesn't seem to resolve it in that case.

ThetaSinner:
> @ThetaSinner Why do you want an AppAgentWs to be reactive? I see that @mattyg did that in mewsfeed, but I cannot see why that's necessary there, either. It seems like the AppAgentWs can just be a variable with let and doesn't need to be reactive.

@jost-s  Yes to what @mattyg said and I'm going to expand a little bit.
- I don't want the app to have to wait to render until the client is connected, all my structural components can render then I can render placeholders for anything that needs data until the client is connected.
- I don't want to have to reload the page if the connection to the back-end drops for any reason. I think it's simpler to reconnect a websocket and resume sending requests than it is to persist and recreate client state on reload.
- I don't want to pass the client through all my components in either React or Vue (I don't know anything about Svelte). In Vue provide/inject is the mechanism for deep passing, in React that would be a hook that provides the client. In either case the provide should be reactive so that the UI can proceed with rendering without waiting for a value.

I'm still experimenting but I've actually already taken this a [step further](https://github.com/ThetaSinner/h-drone-swarm/blob/main/ui/src/client/index.ts#L13) and wrapped the client. This lets me transparently reconnect the client which I can't do with this client because construction/connection is atomic. So I am injecting my client wrapper into the components that need it then they go ahead and render while creating a promise on `waitForConnected` that let's them fill or create in data once a client is available, regardless of which connection that is under the hood.

ThetaSinner:
> I just ran into this issue in another context where I am passing the AppAgentWebsocket into new class as a constructor param, and then trying to listen to signals within that class. Unfortunately this PR doesn't seem to resolve it in that case.

@mattyg any chance you can point me at an example and I'll see if I can figure out why this isn't working everywhere?

mattyg:
> > I just ran into this issue in another context where I am passing the AppAgentWebsocket into new class as a constructor param, and then trying to listen to signals within that class. Unfortunately this PR doesn't seem to resolve it in that case.
> 
> @mattyg any chance you can point me at an example and I'll see if I can figure out why this isn't working everywhere?

Well I tried to make an example, but it worked properly, and then I fixed it in my use case as well, so nevermind!

ThetaSinner:
@mattyg Cool okay, if it pops up again feel free to send it my way :)

jost-s:
@ThetaSinner With the wrapper, do you still need the Emittery workaround?

jost-s:
@ThetaSinner I suppose so. I'll merge this, thanks you for the work!

jost-s (approved):

ThetaSinner:
Sorry I hadn't got to this yet but yes, the Vue proxies are deep so even with a wrapper the `this` will move.

jost-s:
> Sorry I hadn't got to this yet but yes, the Vue proxies are deep so even with a wrapper the `this` will move.

Right, okay. I have been unable to reproduce the problem too, even in a JS file using Vue's `ref` function.

ThetaSinner:
The code I used to test the fix is

```js
import Emittery from 'emittery';

class T extends Emittery {
  constructor() {
    super();

    // // Ensure all super methods are bound to this instance because Emittery relies on `this` being the instance.
    // Object.getOwnPropertyNames(Emittery.prototype).forEach((name) => {
    //     if (typeof this[name] === 'function') {
    //         this[name] = this[name].bind(this);
    //     }
    // });
  }

  doOn(func) {
    console.log(this);
    this.on('hello', func);
  }
}

let t = new Proxy(new T(), {});
t.doOn(() => {
    console.log('got a hello');
});

t.emit('hello');
```

It can just be run with node and the fix can be commented in.

I get the same Emittery error I got from a Vue app

```
file:///home/thetasinner/source/holo/holochain-client-js/node_modules/emittery/index.js:27
        if (!events.has(eventName)) {
```

jost-s:
Thanks for that. Do you have a clue why something like
```javascript
const appAgentWs = await AppAgentWs.connect(...);
const ws = ref(appAgentWs);
ws.value.client.on('signal', () => console.log('signalled'));
ws.value.client.emit('signal', ...);
```
works?

ThetaSinner:
That does fail for me, if I do `npm i -D vue` and create a Javascript file called `test2.js` containing

```
import { AppAgentWebsocket } from './lib/index.js';
import { ref } from 'vue';

const appAgentWs = await AppAgentWebsocket.connect('ws://localhost:8080');
const ws = ref(appAgentWs);
ws.value.on('signal', (m) => console.log('signalled', m));
ws.value.emitter.emit('signal', {
    test: 'message'
});
```

I can then start a sandbox, run it with `hc sandbox run -p 8080` and run that script against it, I get an error

```
file:///home/thetasinner/source/holo/holochain-client-js/node_modules/emittery/index.js:27
        if (!events.has(eventName)) {
                    ^

TypeError: Cannot read properties of undefined (reading 'has')
    at getListeners (file:///home/thetasinner/source/holo/holochain-client-js/node_modules/emittery/index.js:27:14)
    at Proxy.on (file:///home/thetasinner/source/holo/holochain-client-js/node_modules/emittery/index.js:280:14)
    at Proxy.on (file:///home/thetasinner/source/holo/holochain-client-js/lib/api/app-agent/websocket.js:171:29)
    at file:///home/thetasinner/source/holo/holochain-client-js/test2.js:6:10
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

The test script runs fine if I comment my fix back in

PR #205, opened 2023-09-16
Sign zome calls via electron

mattyg:
Support for signing zome calls in electron, as expected by https://github.com/mattyg/holochain-kangaroo-electron ( which itself exposes http://github.com/mattyg/holochain-lair-signer )

If `window.__HC_LAUNCHER_ENV__.FRAMEWORK` is set to 'electron', then the client will try to call `window.electronAPI.signZomeCall()`. If `FRAMEWORK` is set to 'tauri' or undefined, then it will call tauri's `invoke()` function.

This also changes how the client checks if it is in a launcher environment -- instead of checking `window.__HC_LAUNCHER_ENV__` once when initially loaded, it will check every time a launcher overridable function is called.


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## \[Unreleased\]
 
 ### Added
+- Support for signing zome calls in electron via `window.electronAPI.signZomeCall` when `__HC_LAUNCHER_ENV__.ENVIRONMENT == 'elctron'`.
```
jost-s:
```suggestion
- Support for signing zome calls in electron via `window.electronAPI.signZomeCall` when `__HC_LAUNCHER_ENV__.ENVIRONMENT == 'electron'`. 
```

src/api/app/websocket.ts
```diff
@@ -204,9 +204,19 @@ const callZomeTransform: Transformer<
...
+      signedZomeCall = await signZomeCall(request);
+    } else if (env.FRAMEWORK === "electron") {
+      signedZomeCall = await signZomeCallElectron(request);
+    } else if (env.FRAMEWORK === "tauri" || !env.FRAMEWORK) {
+      signedZomeCall = await signZomeCallTauri(request);
+    } else {
+      signedZomeCall = await signZomeCallTauri(request);
+    }
+
     return signedZomeCall;
```
```suggestion
if (!env) {
  return signZomeCall(request);
} else if (env.FRAMEWORK === "electron") {
  return signZomeCallElectron(request);
}
return signZomeCallTauri(request);
```

jost-s (approved):
LGTM
please fix the typo and we could shorten the zome call signing function

PR #208, opened 2023-10-01
build(zome-call-signing): swap hash wasm for native packages

jost-s:


jost-s:
This was just to use the CI pipelines.

Issue #211, opened 2023-11-08
Does not work with holochain v0.3.0-beta-dev.23


guillemcordoba:
While trying to upgrade to that version of holochain, @holochain/client stops working altogether. Any call that it tries to make, results in this kind of error being returned:

```
{
  type: { deserialization: null },
  data: 'Bytes(Deserialize("wrong msgpack marker FixStr(11)"))'
}
```

This happens with the `AdminInterface` as well as with the `AppInterface`. 

Maybe this is not an issue with the client... I've spent a lot of time exploring its causes with no result. I did find [this issue](https://github.com/3Hren/msgpack-rust/issues/336), and also noticed that now the holochain git repository depends on two versions of rmp-serde, 0.15 and 1.1 - maybe that's the cause? But I don't think so because I _was_ able to run sweettests without problems.

Maybe this PR was the one that introduced the problem? https://github.com/holochain/holochain/pull/2844

jost-s:
Yes, the ser/de changed somehow. I'll release a matching client for the time being and look into reverting this change on the Holochain side. It's not for the better IMO, from
```
{
  type: "error",
  data: ...
}
```
to
```
  type: { error: null },
  data: ...
}
```

jost-s:
fixed by #212
tryorama release to follow

PR #212, opened 2023-11-08
Prepare for Holochain 0.3

ThetaSinner:
See this issue #211 and this conversation https://discord.com/channels/727959104777224303/728647693244694648/1171849740850319370

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -188,6 +188,14 @@ export class WsClient extends Emittery {
     resolve: RequestResolver,
     reject: RequestRejecter
   ) {
+    function toHexString(byteArray: number[]) {
+      return Array.from(byteArray, function(byte) {
+        return ('0' + (byte & 0xFF).toString(16)).slice(-2);
+      }).join(', ')
+    }
+
+    console.log("sending message", request, toHexString(encode(request) as unknown as number[]))
+
     const id = this.index;
     const encodedMsg = encode({
       id,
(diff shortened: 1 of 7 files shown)


src/api/client.ts
```diff
@@ -188,6 +188,14 @@ export class WsClient extends Emittery {
     resolve: RequestResolver,
     reject: RequestRejecter
   ) {
+    function toHexString(byteArray: number[]) {
+      return Array.from(byteArray, function(byte) {
+        return ('0' + (byte & 0xFF).toString(16)).slice(-2);
+      }).join(', ')
+    }
+
+    console.log("sending message", request, toHexString(encode(request) as unknown as number[]))
```
ThetaSinner: This is super useful for debugging, do not intend to it to be merged to `main`!

ThetaSinner:
This is enough to get a `list-apps` admin API call working, the tests are completely broken

jost-s:
Yeah groovy! No, I hadn't noticed that the ser/de had changed on the Holochain side. I'm grateful for you help. The tests weren't completely broken, only 2 of them weren't passing. The fix was the same as you did for serializing requests, but for deserializing responses. Should be good to go now.

ThetaSinner:
Oh good! I ran them locally and they dumped out a lot of complaints so I assumed I'd caused some breakage. Great that it was closer than I thought :)

PR #225, opened 2024-01-28
Change .connect() api

mattyg:
It feels very weird that we have to pass a url of `new URL('ws://UNUSED')` to setup a websocket connection within a launcher environment.

This PR:
- makes all api `.connect()` functions take an optional `options` object, which contains optional properties: `url` and `defaultTimeout`
- makes the `connect` function throw an error if the url is not defined by parameter, nor by launcher env
- changes the order of parameters in the `AppAgentWebsocket.connect` function, so the required parameter goes first

This would be a breaking change.


mattyg:
Idk @jost-s this api feels weird too -- then you might be connecting with `AppWebsocket.connect(undefined, 50000)` to set a timeout.

Maybe connect should just take a single options object instead?

mattyg:
Let me know what you think @jost-s -- we might also want to get some feedback first.

jost-s:
Sorry @mattyg for not getting back to you. I don't know what happened, I saw the PR a while ago.

It looks fine to me, I'm happy with your proposal. I'll check with Holo and Components but don't think they'll object.

jost-s (approved):

mattyg:
heads up @jost-s I don't actually have permissions to merge this -- if you were waiting on me

jost-s:
No, I've just been ill.

PR #227, opened 2024-01-30
Add workaround for cell_id being a proxy object in Vue

matthme:
Adds a workaround for the case where the `AppAgentClient` is used as a proxy object in Vue and therefore the `cell_id` is passed as a proxy object to callZome instead of a normal javascript array.

### DIFF
--- a/src/api/app-agent/websocket.ts
+++ b/src/api/app-agent/websocket.ts
@@ -163,7 +163,7 @@ export class AppAgentWebsocket implements AppAgentClient {
       const zomeCallPayload: CallZomeRequest = {
         ...omit(request, "role_name"),
         provenance: this.myPubKey,
-        cell_id,
+        cell_id: [cell_id[0], cell_id[1]],
       };
       return this.appWebsocket.callZome(zomeCallPayload, timeout);
     } else if ("cell_id" in request && request.cell_id) {


matthme:
It's maybe debatable whether the workaround for this should be placed in the js-client but it seems like a negligible change to add it in that place and makes life easier for implementors of host zome call signers because it's quite unexpected to get a proxy object in the zome call request.

jost-s:
I'm fine with merging that, sure. Can you show an example in Vue where that falls over?

matthme:
Yes - you can run the `hc scaffold example forum` (the holochain 0.1 version of it) with Vue as framework and then you will see that zome calls throw the error `Error: An object could not be cloned.`. This is electron that's complaining because it can't serialize the cell_id since it's not an actual javascript object but a Vue proxy object. The latter you can see if you log the cell_id [here](https://github.com/holochain/holochain-client-js/blob/6e443e0bfc500a11aca6435a05ef79507b71bd2d/src/api/app-agent/websocket.ts#L162) in the js-client.

jost-s:
I can not confirm that. I just created a new forum app with Vue, ran npm install and start and it works fine
image


It would be surprising if this hasn't been working all along. Can you doublecheck if this happens for you on a freshly scaffolded project?

matthme:
Oh sorry, my instructions were not correct. I forgot that I ran it with the scaffolding from [this](https://github.com/holochain/scaffolding/pull/200) PR. So you'd need to use hc-spin to start the app instead of hc-launch since in the hc-launch case the js-client does some conversions of the zome call request object before sending it over IPC to the tauri backend.

matthme:
Or in your existing project:
`npm install --save-dev @holochain/hc-spin@">=0.100.0 <0.200.0"`
and then modify the `launch:happ` script to:
```
    "launch:happ": "hc-spin -n $AGENTS --ui-port $UI_PORT workdir/forum.happ",
```

jost-s (approved):
Yes, fine by me. I don't know why the proxy array isn't handled in this particular case, but I don't want to spend time figuring out. This is a completely harmless workaround. Thanks!

jost-s:
released as v0.12.8

PR #235, opened 2024-03-21
Add allowed_origins to AttachAppInterfaceRequest and forwarding websocket connection options

matthme:
Addresses the changes of https://github.com/holochain/holochain/pull/3460

This PR forwards the full Websocket connection options - would it potentially be better to only forward the origin argument?


src/api/common.ts
```diff
@@ -180,6 +181,11 @@ export interface WebsocketConnectionOptions {
    */
   url?: URL;
 
+  /**
+   * Options to pass to the underlying websocket connection.
+   */
+  options?: IsoWebSocket.ClientOptions;
```
ThetaSinner: I'd slightly prefer that we don't expose the types we're using internally and provide a type we've defined here that contains just the options that we intend to expose. Just for the reason of being able to swap out the implementation with minimal interface changes. That's more of a style/maintainability thing though, this is just a type and isn't doing anything functional.

ThetaSinner (approved):
Looks fantastic, thank you for taking care of this!

src/api/common.ts
matthme: Absolutely...so only the origin option for now?

matthme: Okay, changed it to only expose the origin property.

ThetaSinner: Nice yeah that looks good. Looks like the doc needs rebuilding?

matthme: How would I do that? I ran npm run build before commiting which runs npm run build:docs as well...when I run it again locally nothing changes.

jost-s: Yeah, `npm run build` should do it. You are a contributor in this repo, so if you copy the branch to the repo, I'd give it a go too.

matthme: I have no push access to the repo unfortunately, otherwise I would not have made a fork

jost-s: Can you try again? You should have push access.

jost-s: And sorry for not having done this earlier. I misread the permissions.

matthme: Yep, just pushed

jost-s: It's always exported because it's used in an exported object = WsClient. It only works if you assign the origin string a new type and use that in `WsClient.connect`.

matthme: Ah, I see now yes. Unfortunate that the name `WebsocketConnectionOptions` is already in use...maybe we want to rename this.

matthme: Not sure the naming is optimal but the problem should be addressed with the latest commit.

jost-s: Yes, it is.
What do you think of this:
```typescript
/**
 * @public
 */
export type WsClientOptions = Pick<IsoWebSocket.ClientOptions, "origin">;

/**
 * Options for a Websocket connection.
 *
 * @public
 */
export interface WebsocketConnectionOptions {
  /**
   * The `ws://` URL of the Websocket server to connect to. Not required when connecting to App API from a Launcher or Kangaroo environment.
   */
  url?: URL;

  /**
   * Options to pass to the underlying websocket connection.
   */
  wsClientOptions?: WsClientOptions;
```

matthme: Yes I think it's better - changed it.

jost-s: I also meant this
```typescript
/**
 * @public
 */
export type WsClientOptions = Pick<IsoWebSocket.ClientOptions, "origin">;
```

matthme: I am confident that I will eventually get there ;). Gave it another try.

jost-s: So am I! =)

Thanks. I checked out the branch to look for the test errors and saw that all tests would need updating now, because `allowed_origin` becomes a mandatory param. That made me rethink this and I've pushed the suggestion to your branch in this repo. Please have a look at [the commit](https://github.com/holochain/holochain-client-js/commit/35131c333d68772c096babad45862f7e40b76b5e).

jost-s: I updated the existing tests with `allowed_origins` and added a test for this allowed_origins feature. It's failing at the moment because it's not been released on Holochain.

jost-s:
All these changes have been included in #236

PR #240, opened 2024-04-25
App ws auth

ThetaSinner:


test/e2e/index.ts
```diff
@@ -1298,28 +1287,3 @@ test(
     t.equal(response, "hi", "updated coordinator zomes can be called");
   })
 );
-
-test(
```
ThetaSinner, line 1302: Couldn't find a sensible way to retain this. I think if the websocket connection drops then the app needs to reconnect

.github/workflows/integration-test.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         run: nix develop -c $SHELL -c "npm ci"
 
       - name: Run tests
-        run: nix develop -c $SHELL -c "npm t"
+        run: nix develop --override-input versions/holochain "github:holochain/holochain/holochain-0.3.0-beta-dev.48" -c $SHELL -c "npm t"
```
ThetaSinner, line 68: Until Holonix is updated we need to test against a specific Holochain that is newer than the one in weekly.

As we work up to automatically testing new releases before releasing Holonix updates, we should find a way to make this simpler than changing the workflow file... Not sure how!

package.json
```diff
@@ -32,7 +32,8 @@
   ],
   "scripts": {
     "lint": "eslint --fix --ext .ts src test .eslintrc.cjs",
-    "test:app-agent": "RUST_LOG=error RUST_BACKTRACE=1 tsx test/e2e/app-agent-websocket.ts",
+    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
```
jost-s, line 35: 💯

src/api/client.ts
```diff
@@ -159,38 +165,48 @@ export class WsClient extends Emittery {
...
+   * @param request The authentication request, containing an app authentication token.
+   */
+  async authenticate(request: AppAuthenticationRequest): Promise<void> {
+    return this.exchange(request, (request, resolve) => {
+      const encodedMsg = encode({
+        type: "authenticate",
+        data: encode(request),
+      });
+      this.socket.send(encodedMsg);
+      // Message just needs to be send first, no need to wait for a response or even require a flush
```
jost-s:
```suggestion
      // Message just needs to be sent first, no need to wait for a response or even require a flush
```

src/api/common.ts
```diff
@@ -19,7 +19,7 @@ export type Requester<Req, Res> = (req: Req, timeout?: number) => Promise<Res>;
 /**
  * @public
  */
-export type RequesterUnit<Res> = () => Promise<Res>;
+export type RequesterNoArg<Res> = (timeout?: number) => Promise<Res>;
```
jost-s, line 22: Is this a rename from Unit (Rust) to NoArg (JS)? Now it has an arg ;-) Anyhow, not that important

src/api/app/types.ts
```diff
@@ -194,12 +262,26 @@ export type NetworkInfoResponse = NetworkInfo[];
...
  */
-export interface AppApi {
-  appInfo: Requester<AppInfoRequest, AppInfoResponse>;
-  callZome: Requester<CallZomeRequest, CallZomeResponse>;
-  enableCloneCell: Requester<EnableCloneCellRequest, EnableCloneCellResponse>;
-  disableCloneCell: Requester<
-    DisableCloneCellRequest,
-    DisableCloneCellResponse
-  >;
+export interface AppAgentClient {
```
jost-s: I think we can rename all `AppAgent*` to `App*`. WDYT?

src/api/app/types.ts
```diff
@@ -1,22 +1,102 @@
...
+ */
+export type RoleNameCallZomeRequestSigned = Omit<
+  CallZomeRequestSigned,
+  "cell_id"
+> & { role_name: RoleName };
+
+/**
+ * @public
+ */
+export type AppAgentCallZomeRequest =
```
jost-s: rename here too?

jost-s (approved):
Groovy!

src/api/common.ts
ThetaSinner: Thanks diff tool! `RequesterUnit` wasn't used so I deleted it. This one is supposed to be "no request argument"

src/api/app/types.ts
ThetaSinner: Heck, had moved this stuff without looking too closely - hadn't realised how many instances there were of 'agent' still. On it!

PR #242, opened 2024-04-26
Adapt to other API changes for app ws security

ThetaSinner:


test/e2e/index.ts
```diff
@@ -920,6 +920,9 @@ test(
 
     interfaces = await admin.listAppInterfaces();
     t.equal(interfaces.length, 1);
+    t.true(interfaces[0].port > 0);
```
ThetaSinner, line 923: Test wasn't really checking the response

PR #243, opened 2024-04-26
Appwebsocket authentication token from launcher env

matthme:
* takes the app websocket token from the launcher env if available
* removes deprecated framework specific zome call signing logic


matthme:
converted it to a draft because I haven't actually tested it in practice yet

src/api/app/websocket.ts
```diff
@@ -161,10 +159,7 @@ export class AppWebsocket implements AppClient {
    * @param options - {@link (WebsocketConnectionOptions:interface)}
    * @returns A new instance of an AppWebsocket.
    */
-  static async connect(
-    token: AppAuthenticationToken,
-    options: WebsocketConnectionOptions = {}
-  ) {
+  static async connect(options: AppWebsocketConnectionOptions = {}) {
```
ThetaSinner, line 162: Changelog entry for this please

src/api/app/websocket.ts
```diff
@@ -180,7 +175,17 @@ export class AppWebsocket implements AppClient {
     }
 
     const client = await WsClient.connect(options.url, options.wsClientOptions);
-    await client.authenticate({ token });
+    if (env?.APP_INTERFACE_TOKEN) {
+      await client.authenticate({ token: env.APP_INTERFACE_TOKEN });
```
ThetaSinner, line 181: Maybe a comment here to note that the launcher must provide a `single_use = false` token?

src/api/app/websocket.ts
```diff
@@ -426,14 +431,7 @@ const callZomeTransform: Transformer<
...
     } else {
-      const env = getLauncherEnvironment();
-      if (!env) {
-        return signZomeCall(request);
-      }
-      if (env.FRAMEWORK === "electron") {
-        return signZomeCallElectron(request);
-      }
-      return signZomeCallTauri(request);
+      return signZomeCall(request);
```
ThetaSinner, line 436: Why is zome call singing by the runtime being removed?

src/api/app/websocket.ts
ThetaSinner: Ahh, done while I was reviewing

src/api/app/websocket.ts
matthme: Only the legacy methods are removed. The framework agnostic `hostSigner.signZomeCall()` is still there.

src/api/app/websocket.ts
matthme: Hmm...good point. Do you see any problems with launcher providing `single_use = false` tokens?

src/api/app/websocket.ts
ThetaSinner: Ahh I see what you're saying, nice okay!

ThetaSinner (approved):

src/api/app/websocket.ts
ThetaSinner: The alternative would be providing a host call to retrieve a token each time an appws is created. I think `single_use = false` is fine to start and maybe we can consider retrieving tokens on demand as an enhancement later?

PR #244, opened 2024-04-26
Fix bad dir imports

ThetaSinner:
My IDE thought (wrongly) that because the project uses TypeScript, that dir imports are valid. They are not once the library is built and running as a JS dependency!

I've updated CI with a check that would have caught this mistake.

### DIFF
--- a/.github/workflows/bundling-test.yml
+++ b/.github/workflows/bundling-test.yml
@@ -34,6 +34,9 @@ jobs:
       - name: Build client library
         run: npm run build
 
+      - name: Ensure Node.js can load the built client library
+        run: node lib/index.js
+
       - name: Create package tarball
         run: npm pack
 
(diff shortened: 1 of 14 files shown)


.github/workflows/bundling-test.yml
```diff
@@ -34,6 +34,9 @@ jobs:
       - name: Build client library
         run: npm run build
 
+      - name: Ensure Node.js can load the built client library
+        run: node lib/index.js
```
ThetaSinner, line 38: This is enough to check that module resolution is working in a plain Node+Js environment

src/api/app/types.ts
```diff
@@ -285,6 +285,9 @@ export interface AppClient {
   networkInfo(args: AppNetworkInfoRequest): Promise<NetworkInfoResponse>;
 }
 
+/**
+ * @public
+ */
```
ThetaSinner, line 290: Fixes a newly introduced warning

jost-s:
Right, yes, with ES modules, importing the .js file is the universal way for JS and TS.

PR #247, opened 2024-05-09
Add optional callZomeTransform

matthme:
Until recently it was possible to overwrite the zome call signing logic [like this](https://github.com/lightningrodlabs/we/blob/7bbdf403b6cffdedafbae66f4d231176cfe67ca3/utils/applet-iframe/src/index.ts#L337). But this doesn't seem to be possible anymore since there is now a `callZomeRequester` that's private and readonly. This PR allows to choose a custom zome call transform for use cases that the default CallZomeTransform cannot handle.


src/api/app/websocket.ts
```diff
@@ -92,11 +100,15 @@ export class AppWebsocket implements AppClient {
   private constructor(
     client: WsClient,
     appInfo: AppInfo,
-    defaultTimeout?: number
+    defaultTimeout?: number,
+    callZomeTransform?: CallZomeTransform
```
ThetaSinner: Which are users more likely to set? I'd say the `callZomeTransform` is probably more common than overriding the timeout so maybe that comes before the `defaultTimeout` in the params list?

ThetaSinner: Probably not so important either way

ThetaSinner (approved):
Nice, looks good! One question that might not need looking at

src/api/app/websocket.ts
matthme: Good question, I was assuming the opposite but you might be right. Holochain has an internal timeout of 60 seconds is that right? So the only sensible reason to change the timeout here would be to make it shorter which I'm not sure many people would want to do.

src/api/app/websocket.ts
```diff
@@ -92,11 +100,15 @@ export class AppWebsocket implements AppClient {
...
   ) {
     this.client = client;
     this.myPubKey = appInfo.agent_pub_key;
     this.defaultTimeout = defaultTimeout ?? DEFAULT_TIMEOUT;
+    this.callZomeTransform = callZomeTransform
```
```suggestion
    this.callZomeTransform = callZomeTransform ?? defaultCallZomeTransform
```

jost-s (changes requested):
Teeny change request

src/api/app/websocket.ts
matthme: done.

jost-s (approved):

PR #259, opened 2024-06-27
Use @spartan-hc/holo-hash for managing holo-hashes

mjbrisebois:
Here is a quick example of how we could start using a library to manage holo-hashes while preserving legacy support.

Most of the [helper functions](https://github.com/holochain/holochain-client-js/tree/develop/src/utils) in utils could be replaced by the native methods on [`HoloHash`](https://github.com/spartan-holochain-counsel/holo-hash-js/blob/master/src/index.ts#L106) instances.

- [`sliceDhtLocation`](https://github.com/holochain/holochain-client-js/blob/develop/src/utils/hash-parts.ts#L29) => [`<HoloHash>.getDHTLocation`](https://github.com/spartan-holochain-counsel/holo-hash-js/blob/master/src/index.ts#L306)
- [`sliceCore32`](https://github.com/holochain/holochain-client-js/blob/develop/src/utils/hash-parts.ts#L45) => [`<HoloHash>.getHash`](https://github.com/spartan-holochain-counsel/holo-hash-js/blob/master/src/index.ts#L285)
- [`sliceHashType`](https://github.com/holochain/holochain-client-js/blob/develop/src/utils/hash-parts.ts#L61) => [`<HoloHash>.getPrefix`](https://github.com/spartan-holochain-counsel/holo-hash-js/blob/master/src/index.ts#L268)
- etc...


mjbrisebois:
@jost-s do you see value in this?

jost-s (commented):
Yes, I do see value in this. Looks good to me

src/utils/base64.ts
```diff
@@ -21,6 +22,8 @@ export function decodeHashFromBase64(hash: HoloHashB64): HoloHash {
  *
  * @public
  */
-export function encodeHashToBase64(hash: HoloHash): HoloHashB64 {
-  return `u${Base64.fromUint8Array(hash, true)}`;
+export function encodeHashToBase64(hash: Uint8Array): HoloHashB64 {
+  return String(new HoloHash( hash ));
 }
+
+export * from "@spartan-hc/holo-hash";
```
mjbrisebois: This causes ambiguity because the `HoloHash` classes have the same name as all the `types.ts` exports.  How should we export the class implementations without conflict?

jost-s: I think we should ship this client with only one type for each type. So I wouldn't want to export a legacy type _and_ the new type under an alias. Therefore I'd replace the existing type with the spartan-hc one.

What do you think about this, @ThetaSinner?

mjbrisebois: Supporting both `HoloHash` and `Uint8Array` input for most functions is important so that users are not forced to wrap all input in `new HoloHash( input )`.  But I don't know enough about typescript to do that.  Would it work to extend all the types with an "or-pipe"?

Example
```ts
export type AgentPubKey = Uint8Array | HoloHashLib.AgentPubKey;
```

mjbrisebois: Nevermind, that doesn't solve the duplicate export issue.

jost-s: Just to let you know this is still on my radar.

src/api/zome-call-signing.ts
```diff
@@ -54,24 +55,18 @@ export const setSigningCredentials = (
...
  * Generates a key pair for signing zome calls.
  *
- * @param agentPubKey - The agent pub key to take 4 last bytes (= DHT location)
- * from (optional).
  * @returns The signing key pair and an agent pub key based on the public key.
  *
  * @public
  */
-export const generateSigningKeyPair: (
-  agentPubKey?: AgentPubKey
```
mjbrisebois, line 64: What is the purpose of forcing the same location bytes?  Imposing location bytes on the end of random bytes makes the DHT location check fail.

src/utils/fake-hash.ts
```diff
@@ -1,77 +1,77 @@
 import { range } from "lodash-es";
 import { randomByteArray } from "../api/zome-call-signing.js";
 import { DnaHash, ActionHash, AgentPubKey, EntryHash } from "../types.js";
-import { dhtLocationFrom32 } from "./hash-parts.js";
-
-async function fakeValidHash<T extends Uint8Array>(
```
mjbrisebois, line 6: Since this function is not exported, it is safe to assume it was only used by this file.  Correct?

src/api/zome-call-signing.ts
jost-s: There was no purpose, it was just a quick way of setting location bytes.

src/utils/hash-parts.ts
```diff
@@ -66,11 +70,13 @@ export function sliceCore32(
  * @public
  */
 export function sliceDhtLocation(
-  hash: AgentPubKey | EntryHash | ActionHash
+  hash: HoloHash | Uint8Array,
 ): Uint8Array {
+  if (!(hash instanceof HoloHash))
```
jost-s:
```suggestion
  if (hash instanceof Uint8Array)
```

jost-s (approved):
Yes, that's the way to go, use `HoloHash | Uint8Array` for input to maintain compatibility.

Other than that, the transpiler didn't complain and everything seems to work fine.

mjbrisebois:
Ready to merge then?

jost-s:
Yes, thanks for starting this off!

mattyg:
This introduced some problems: by using the classes in `@spartan-hc/holo-hash` for type definitions it makes it appear that the types you are getting back from zome calls are instances of those classes, but they're not.

I.e. if my zome call returns an `AgentPubKey`, then the js client is receiving a Uint8Array, not a `@spartan-hc/holo-hash` AgentPubKey, so its mis-typed. Compound types using these types are also mis-typed, and the return values from other client calls (not zome calls) are also mis-typed.

Issue #265, opened 2024-07-31
Complete integration of `@hc-spartan/holo-hash` or revert


mattyg:
This PR #259 introduced some bugs: by using the classes in `@spartan-hc/holo-hash` for type definitions it makes it appear that the types you are getting back from client calls are instances of those classes, but they're not.

We either need to modify client responses to transform Uint8Arrays into their expected `@spartan-hc/holo-hash` class instances, or to revert.

matthme:
> We either need to modify client responses to transform Uint8Arrays into their expected @spartan-hc/holo-hash class instances

I would advocate against using objects in client responses as that will create interoperability issues if someone is using a different js-client. I already noticed that the latest PR breaks apps using that new client within Moss which is using an older version of the client because Moss expects pure Uint8Arrays in iframe messages from apps but it gets javascript objects instead.

matthme:
I think it would be smart to stay with the pure serialized responses as anyone would get them when decoding websocket responses from javascript and keep these classes as optional sugar to make things more ergonomic in other contexts.

guillemcordoba:
I agree... Right now I think the PR makes the types for the responses coming from the `AdminWebsocket` just incorrect, as typescript would need the `HoloHash` classes to compile, but then when the responses come they are just `Uint8Array`s in reality. That could be addressed in the transformer for the `AdminWebsocket` to convert `Uint8Arrays` into `HoloHash`es. However, the core @holochain/client package can't do things like go over all response types that the app call zomes return, and a bunch of them will be `Uint8Array`s. So if this change stays, the app developers would be left with a weird mixture of `HoloHash`es and `Uint8Array`, and it may not be clear to them when they are receiving what, and also it's more confusing because at the conceptual level they are equivalent. 

I think for this to be properly implemented it should be at the level of msgpack's [extension types](https://github.com/msgpack/msgpack/blob/master/spec.md#extension-types), where one of the extension types would be  `HoloHash`. I'm not sure exactly how this would need to be implemented, but I think it also needs work from the holochain core side to mark `HoloHashes` as special rmp-serde types. If done this way, I would support it wholeheartedly, and could even contribute to it. The biggest win for me here is that we could declare the `equals` function of `HoloHash`es as equality by value and not by reference, and then `hash1 === hash2` would just work as expected (right now it doesn't, we need to do weird things like `hash1.toString() === hash2.toString()`.

jost-s:
[...]

PR #266, opened 2024-07-31
Mistyped client call responses

mattyg:
Failing test demonstrating #265

### DIFF
--- a/src/api/app/websocket.ts
+++ b/src/api/app/websocket.ts
@@ -231,6 +231,7 @@ export class AppWebsocket implements AppClient {
         `The app your connection token was issued for was not found. The app needs to be installed and enabled.`
       );
     }
+    appInfo.agent_pub_key = new AgentPubKey(appInfo.agent_pub_key);
 
     return new AppWebsocket(
       client,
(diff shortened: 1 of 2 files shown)


mattyg:
Heads up this is just a demonstration with one client call -- appInfo. The same issue needs to be resolved for all other client calls that take in hash types in their input type or have hash types in their responses.

For responses we have a simple solution: transform the response type and cast any Uint8Arrays representing hashes into a HoloHash instance.

For requests I don't know what the best solution is: do we want to require the input type to actually be HoloHash instances? Or do we want another set of types for when hashes are used in inputs, or type them all Uint8Array?

mjbrisebois:
I think all inputs should accept `Uint8Array` and then it can be transformed to the expected hash type like the example in [utils/hash-parts.ts](https://github.com/holochain/holochain-client-js/blob/develop/src/utils/hash-parts.ts#L34-L35).

As for responses, I don't have a strong opinion.  However, I would lean towards the client not automatically transforming the payloads.

jost-s:
I agree with both, inputs should accept `Uint8Array` and responses remain unaltered.
Perhaps inputs should actually accept both `Uint8Array` and `HoloHash` where applicable.

Would it be helpful to introduce another type which is `Uint8Array | HoloHash`? Actually I realize I need to discuss this in the office hours call next week. This change has wider implications than I saw.

mattyg:
In that case why dont we just revert to using the original types? We can still use the @spartan-hc/holo-hash utility functiom implementations under-the-hood

jost-s:
Reverted the spartan-hc changes.

PR #272, opened 2024-08-30
improve Signal type

c12i:
This PR attempts to address an issue with the `Signal` type that was causing lack of intellisense and type errors when accessing properties like `signal.App` of the `SignalCb` type.

```ts
client?.on("signal", (signal) {
    console.log(signal.App.zome_name) // Property 'App' does not exist on type 'Signal'. Property 'App' does not exist on type '{ System: SystemSignal; }'
 });
```

### DIFF
--- a/src/api/app/types.ts
+++ b/src/api/app/types.ts
@@ -247,24 +247,21 @@ export const SignalType = {
 /**
  * @public
  */
-export type RawSignal =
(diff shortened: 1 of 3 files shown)


c12i:
Closing since this PR opts out of the type safety of the initial implementation

Issue #271, opened 2024-08-27
Automatically try re-establishing connection if socket is closed


matthme:
If sockets close, e.g. due to sleep mode of one's laptop, it would be nice if the js-client would automatically attempt to re-establish the connection with the connection options as supplied for the initial connection before [throwing](https://github.com/holochain/holochain-client-js/blob/16aa191c04974351377426c7f0ffd9f8893779d4/src/api/client.ts#L216) an `Error: Socket is not open`.

Apparently this logic was already implemented at some point but [got removed](https://github.com/holochain/holochain-client-js/commit/8253f559f20f247c4cefaf32bb4d86882766b8ae#diff-b8947d6beda79d7ac4e7fbaf1da6b81f496c35e8f87635fc0b15599ee8dd0478L209-L228) a while ago.

jost-s:
fixed by #193

PR #279, opened 2024-10-14
Add logic back to try to reconnect to the app websocket if socket is closed

matthme:
Addresses #271

### DIFF
--- a/src/api/app/websocket.ts
+++ b/src/api/app/websocket.ts
@@ -449,27 +449,6 @@ export class AppWebsocket implements AppClient {
       transformer
     );
   }
-
-  private containsCell(cellId: CellId) {
-    const appInfo = this.cachedAppInfo;
-    if (!appInfo) {
-      return false;
-    }
-    for (const roleName of Object.keys(appInfo.cell_info)) {
-      for (const cellInfo of appInfo.cell_info[roleName]) {
-        const currentCellId =
-          CellType.Provisioned in cellInfo
-            ? cellInfo[CellType.Provisioned].cell_id
-            : CellType.Cloned in cellInfo
-            ? cellInfo[CellType.Cloned].cell_id
-            : undefined;
-        if (currentCellId && isSameCell(currentCellId, cellId)) {
-          return true;
-        }
-      }
-    }
-    return false;
-  }
 }
 
 const defaultCallZomeTransform: Transformer<
--- a/src/api/app/websocket.ts
+++ b/src/api/app/websocket.ts
@@ -1,7 +1,12 @@
 import Emittery, { UnsubscribeFunction } from "emittery";
 import { omit } from "lodash-es";
 import { AgentPubKey, CellId, InstalledAppId, RoleName } from "../../types.js";
-import { AppInfo, CellType, MemproofMap } from "../admin/index.js";
+import {
+  AppAuthenticationToken,
+  AppInfo,
+  CellType,
+  MemproofMap,
+} from "../admin/index.js";
 import {
   catchError,
   DEFAULT_TIMEOUT,
@@ -77,6 +82,7 @@ export class AppWebsocket implements AppClient {
     CallZomeResponseGeneric<Uint8Array>,
     CallZomeResponse
   >;
+  private readonly appAuthenticationToken: AppAuthenticationToken;
 
   cachedAppInfo?: AppInfo | null;
 
@@ -110,6 +116,7 @@ export class AppWebsocket implements AppClient {
   private constructor(
     client: WsClient,
     appInfo: AppInfo,
+    token: AppAuthenticationToken,
     callZomeTransform?: CallZomeTransform,
     defaultTimeout?: number
   ) {
@@ -118,6 +125,7 @@ export class AppWebsocket implements AppClient {
     this.installedAppId = appInfo.installed_app_id;
     this.defaultTimeout = defaultTimeout ?? DEFAULT_TIMEOUT;
     this.callZomeTransform = callZomeTransform ?? defaultCallZomeTransform;
+    this.appAuthenticationToken = token;
     this.emitter = new Emittery<AppEvents>();
     this.cachedAppInfo = appInfo;
 
@@ -204,18 +212,15 @@ export class AppWebsocket implements AppClient {
 
     const client = await WsClient.connect(options.url, options.wsClientOptions);
 
-    if (env?.APP_INTERFACE_TOKEN) {
-      // Note: This will only work for multiple connections if a single_use = false token is provided
-      await client.authenticate({ token: env.APP_INTERFACE_TOKEN });
-    } else {
-      if (!options.token) {
-        throw new HolochainError(
-          "AppAuthenticationTokenMissing",
-          `unable to connect to Conductor API - no app authentication token provided.`
-        );
-      }
-      await client.authenticate({ token: options.token });
-    }
+    const token = options.token ? options.token : env?.APP_INTERFACE_TOKEN;
+
+    if (!token)
+      throw new HolochainError(
+        "AppAuthenticationTokenMissing",
+        `unable to connect to Conductor API - no app authentication token provided.`
+      );
+
+    await client.authenticate({ token });
 
     const appInfo = await (
       AppWebsocket.requester(client, "app_info", DEFAULT_TIMEOUT) as Requester<
@@ -233,6 +238,7 @@ export class AppWebsocket implements AppClient {
     return new AppWebsocket(
       client,
       appInfo,
+      token,
       options.callZomeTransform,
       options.defaultTimeout
     );
--- a/src/api/app/websocket.ts
+++ b/src/api/app/websocket.ts
@@ -212,7 +212,7 @@ export class AppWebsocket implements AppClient {
 
     const client = await WsClient.connect(options.url, options.wsClientOptions);
 
-    const token = options.token ? options.token : env?.APP_INTERFACE_TOKEN;
+    const token = options.token ?? env?.APP_INTERFACE_TOKEN;
 
     if (!token)
       throw new HolochainError(
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -37,7 +37,7 @@ export interface AppAuthenticationRequest {
 export class WsClient extends Emittery {
   socket: IsoWebSocket;
   url: URL | undefined;
-  options: WsClientOptions;
+  options: WsClientOptions | undefined;
   private pendingRequests: Record<number, HolochainRequest>;
   private index: number;
   private authenticationToken: AppAuthenticationToken | undefined;
@@ -46,7 +46,7 @@ export class WsClient extends Emittery {
     super();
     this.socket = socket;
     this.url = url;
-    this.options = options || {};
+    this.options = options;
     this.pendingRequests = {};
     this.index = 0;
 
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -37,7 +37,7 @@ export interface AppAuthenticationRequest {
 export class WsClient extends Emittery {
   socket: IsoWebSocket;
   url: URL | undefined;
-  options: WsClientOptions | undefined;
+  options: WsClientOptions;
   private pendingRequests: Record<number, HolochainRequest>;
   private index: number;
   private authenticationToken: AppAuthenticationToken | undefined;
@@ -48,7 +48,7 @@ export class WsClient extends Emittery {
     this.registerCloseListener(socket);
     this.socket = socket;
     this.url = url;
-    this.options = options;
+    this.options = options || {};
     this.pendingRequests = {};
     this.index = 0;
   }
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -108,6 +108,7 @@ export class WsClient extends Emittery {
       const invalidTokenCloseListener = (
         closeEvent: IsoWebSocket.CloseEvent
       ) => {
+        this.authenticationToken = undefined;
         reject(
           new HolochainError(
             "InvalidTokenError",
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -44,17 +44,164 @@ export class WsClient extends Emittery {
 
   constructor(socket: IsoWebSocket, url?: URL, options?: WsClientOptions) {
     super();
+    this.registerMessageListener(socket);
+    this.registerCloseListener(socket);
     this.socket = socket;
     this.url = url;
     this.options = options;
     this.pendingRequests = {};
     this.index = 0;
-
-    this.setupSocket();
   }
 
-  private setupSocket() {
-    this.socket.onmessage = async (serializedMessage) => {
+  /**
+   * Instance factory for creating WsClients.
+   *
+   * @param url - The WebSocket URL to connect to.
+   * @param options - Options for the WsClient.
+   * @returns An new instance of the WsClient.
+   */
+  static connect(url: URL, options?: WsClientOptions) {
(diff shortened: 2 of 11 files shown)


src/api/app/websocket.ts
```diff
@@ -204,18 +212,15 @@ export class AppWebsocket implements AppClient {
...
-    } else {
-      if (!options.token) {
-        throw new HolochainError(
-          "AppAuthenticationTokenMissing",
-          `unable to connect to Conductor API - no app authentication token provided.`
-        );
-      }
-      await client.authenticate({ token: options.token });
-    }
+    const token = options.token ? options.token : env?.APP_INTERFACE_TOKEN;
```
```suggestion
    const token = options.token ?? env?.APP_INTERFACE_TOKEN;
```

src/api/client.ts
```diff
@@ -212,6 +214,33 @@ export class WsClient extends Emittery {
         sendHandler(request, resolve, reject);
       });
       return promise as Promise<Response>;
+    } else if (this.url && this.authenticationToken) {
+      const response = new Promise<unknown>((resolve, reject) => {
+        // typescript forgets in this promise scope that this.url is undefined
```
jost-s: I know I wrote this, but it didn't make sense =D
```suggestion
        // typescript forgets in this promise scope that `this.url` is not undefined
```

test/e2e/common.ts
```diff
@@ -127,6 +128,8 @@ export const installAppAndDna = async (
   });
   const issued = await admin.issueAppAuthenticationToken({
     installed_app_id,
+    single_use: nonExpiringToken ? false : true,
```
jost-s:
```suggestion
    single_use: !nonExpiringToken,
```

jost-s (approved):
Minor changes, please. LGTM, thanks!

src/api/client.ts
matthme: I seem to struggle with double negations :)

matthme: You wrote it correctly and I muddled it up.

src/api/client.ts
```diff
@@ -212,6 +214,33 @@ export class WsClient extends Emittery {
         sendHandler(request, resolve, reject);
       });
       return promise as Promise<Response>;
+    } else if (this.url && this.authenticationToken) {
+      const response = new Promise<unknown>((resolve, reject) => {
+        // typescript forgets in this promise scope that `this.url` is not undefined
+        const socket = new IsoWebSocket(this.url as URL, this.options);
+        this.socket = socket;
+        socket.onerror = (errorEvent) => {
```
ThetaSinner: Can this tell the difference between an invalid token and other kinds of error?

It seems like the user experience wouldn't be great if requests always result in a reconnect that fails when a temporary token was in use. If we can tell the token is invalid here, it would be good to clear it so that the client stops trying to reconnect

test/e2e/index.ts
```diff
@@ -1441,6 +1441,52 @@ test(
...
+      zome_name: TEST_ZOME_NAME,
+      fn_name: "bar",
+      provenance: cell_id[1],
+      payload: null,
+    });
+    try {
+      await call;
+      t.pass("websocket was reconnected successfully");
+    } catch (error) {
+      t.fail("websocket was not reconnected");
```
ThetaSinner: Include the error in the fail here so we can see why it failed if it does

src/api/client.ts
matthme: With an invalid token what happens is that you get the "Client disconnected with pending requests" error and this onError callback here is never called in the first place.

test/e2e/index.ts
```diff
@@ -1441,6 +1441,52 @@ test(
...
+      fn_name: "bar",
+      provenance: cell_id[1],
+      payload: null,
+    });
+    try {
+      await call;
+      t.fail(
+        "reconnecting to websocket should have failed due to an invalid token."
+      );
+    } catch (error) {
```
ThetaSinner, line 1513: assert the failure reason here?

src/api/client.ts
matthme: I don't actually understand where the "client disconnected with pending requests" occurs...but if I log something in that `onError` callback it's not getting logged when it fails due to an invalid token.

test/e2e/index.ts
matthme: As written above it throws the "client disconnected with pending requests" error and it seemed to me like it doesn't make sense to match on that.

test/e2e/common.ts
```diff
@@ -97,7 +97,8 @@ export const withConductor =
   };
 
 export const installAppAndDna = async (
-  adminPort: number
+  adminPort: number,
+  nonExpiringToken?: boolean
```
ThetaSinner: Would it make sense to call this parameter `singleUseToken` to stay close to the conductor API naming?

matthme: And then have it be true by default? We can do that yes. It's single-use *and* expiring...I could also make two parameters for expiry and single use.

test/e2e/index.ts
```diff
@@ -1441,6 +1441,52 @@ test(
...
+      await call;
+      t.pass("websocket was reconnected successfully");
+    } catch (error) {
+      t.fail("websocket was not reconnected");
+    }
+  })
+);
+
+test(
+  "client fails to reconnect to websocket if closed before making a zome call and does not end up in a loop",
```
ThetaSinner: This isn't quite testing what I was concerned about with looping. It's repeatedly reconnecting with an invalid token that I was concerned about. If the caller isn't careful about which error is which then they might treat this error as retryable and keep trying to reconnect. I'd prefer if this test could make two zome calls and the second one got a different error message back asking the caller to provide new credentials

matthme: Well, configuring the message to ask for new credentials doesn't work I think due to the "client closed with pending requests" error that doesn't even let us define the error message here.

matthme: Maybe I can look into handling this close event better in the first place.

matthme: But I'm not sure whether we should conflate this with this PR

matthme: There is no way of modifying the error in a way to ask the user to provide a valid token. Any zome call with an invalid token will now lead to the following error:
```
ClientClosedWithPendingRequests: client closed with pending requests - close event code: 1005, request id: 1
```
And this seems to be a result of how the conductor handles things and cannot be affected at the js side as I understand it. This error however is distinguishable by a caller from a "normal" zome call error if required so I don't think there's more we can do here without touching the conductor behavior.

test/e2e/common.ts
matthme: I changed it to take two separate parameters now to be aligned with the conductor API naming: https://github.com/holochain/holochain-client-js/pull/279/commits/d6a7db7113ba2f14a688045a36866e0a83fbf9de

src/api/client.ts
ThetaSinner: Interesting. Well Holochain won't send back anything to tell the client that the connection is going to be closed, it'll just close the connection from its end. Where is that error thrown, can/should it be caught?

test/e2e/index.ts
ThetaSinner: Let's figure out the conversation above first. But I think it's always worth asserting the error message to make sure it doesn't become some array access out of bounds error. We need to know that the test is doing what it claims to do

test/e2e/common.ts
ThetaSinner: That works yes, just expose the API with defaults. I was thinking that single use could have a timeout and when it's false then it could have no expiry. I wouldn't be opposed to the client exposing the API that way because those are the two expected configurations really

ThetaSinner: Though it's probably useful to let people just have access to these

src/api/client.ts
jost-s: I confirm that his happens. It comes in as a `close` event on the websocket with code [`1005`](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code#value) which means "No status received". It's indistinguishable from another unexpected close event which usually is 1005 or 1006. For this reconnect we could register another close handler that unsets the token, but it's not 100 % certain that the close was due to an invalid token.

test/e2e/index.ts
matthme: Okay, I will assert that it's the client closed with pending requests then.

matthme: 6dcaeec4305cd76be0200ff9657f28189063b1e8

src/api/client.ts
jost-s: There's a specific `InvalidTokenError` now which only happens on failed automatic websocket reconnect. If the reconnect fails, the stored token is deleted. No risk of this happening in a loop, because the reconnects are only attempted as part of a request being made by the client consumer.

src/api/client.ts
```diff
@@ -142,18 +63,22 @@ export class WsClient extends Emittery {
   static connect(url: URL, options?: WsClientOptions) {
     return new Promise<WsClient>((resolve, reject) => {
       const socket = new IsoWebSocket(url, options);
-      socket.onerror = (errorEvent) => {
+      socket.addEventListener("error", (errorEvent) => {
```
jost-s, line 66: `addEventListener` used throughout now, with one-off listeners for all calls returning a promise.

src/api/client.ts
```diff
@@ -178,17 +103,49 @@ export class WsClient extends Emittery {
...
+      });
       const encodedMsg = encode({
         type: "authenticate",
         data: encode(request),
       });
       this.socket.send(encodedMsg);
-      // Message just needs to be sent first, no need to wait for a response or even require a flush
-      resolve(null);
+      // Wait before resolving in case authentication fails.
+      setTimeout(() => {
```
jost-s, line 128: If authentication fails, it'll still return a `resolve`. Waiting here prevents that. It's not a universal fool-proof way, ideally a ping can be sent after authenticating or Holochain returns an error or closes the websocket with a specific error code.

src/api/client.ts
```diff
@@ -233,6 +200,144 @@ export class WsClient extends Emittery {
     this.index += 1;
   }
 
+  private registerMessageListener(socket: IsoWebSocket) {
```
jost-s, line 204: Unchanged, just separated into a function and moved within the file.

src/api/client.ts
```diff
@@ -233,6 +200,144 @@ export class WsClient extends Emittery {
...
+        );
+      };
+      this.socket.addEventListener("close", invalidTokenCloseListener, {
+        once: true,
+      });
+
+      this.socket.addEventListener(
+        "open",
+        async (_) => {
+          const encodedMsg = encode({
```
jost-s, line 326: I would have loved to call `authenticate` again here, but that's so convoluted with the `sendHandler` parameter and the `requesterTransformer` and `_request` and `promiseTimeout`, it's head-wrecking. Would be brilliant to simplify that part of the client.

test/e2e/index.ts
```diff
@@ -422,6 +422,28 @@ test(
   })
 );
 
+test(
```
jost-s, line 425: Test unhappy websocket authentication path.

test/e2e/index.ts
```diff
@@ -1441,6 +1461,82 @@ test(
...
+      t.fail(
+        "reconnecting to websocket should have failed due to an invalid token."
+      );
+    } catch (error) {
+      t.assert(
+        error instanceof HolochainError,
+        "error should be of type HolochainError"
+      );
+      assert(error instanceof HolochainError);
+      t.equal(error.name, "InvalidTokenError", "expected an InvalidTokenError");
```
jost-s, line 1519: For the future - create an enum with all possible error kinds https://github.com/holochain/holochain-client-js/issues/281

src/api/common.ts
```diff
@@ -96,6 +96,7 @@ export const promiseTimeout = (
         return res(a);
       })
       .catch((e) => {
+        clearTimeout(id);
```
jost-s, line 99: Without clearing the timeout, failed requests were waiting for the duration of the timeout (60 seconds) before returning.

src/api/client.ts
matthme: Great stuff, thanks for polishing this :+1: . This idea did not occur to me.

src/api/app/websocket.ts
```diff
@@ -77,6 +82,7 @@ export class AppWebsocket implements AppClient {
     CallZomeResponseGeneric<Uint8Array>,
     CallZomeResponse
   >;
+  private readonly appAuthenticationToken: AppAuthenticationToken;
```
ThetaSinner, line 85: I don't believe this is very high risk, but I don't feel comfortable not pointing this out. Storing this token may expose it to an injected script. If it is exfiltrated then it is re-usable elsewhere. 

I much prefer the single-use tokens but I understand that this is being done for UX.

src/api/client.ts
```diff
@@ -178,17 +103,49 @@ export class WsClient extends Emittery {
    * @param request - The authentication request, containing an app authentication token.
    */
   async authenticate(request: AppAuthenticationRequest): Promise<void> {
-    return this.exchange(request, (request, resolve) => {
+    this.authenticationToken = request.token;
+    return this.exchange(request, (request, resolve, reject) => {
+      const invalidTokenCloseListener = (
+        closeEvent: IsoWebSocket.CloseEvent
+      ) => {
+        reject(
```
ThetaSinner, line 112: If we're rejecting because of an invalid token, should `this.authenticationToken` not be cleared here?

test/e2e/index.ts
```diff
@@ -422,6 +422,28 @@ test(
...
+        url: new URL(`ws://localhost:${port}`),
+        wsClientOptions: { origin: "client-test-app" },
+        token: [0],
+      });
+      t.fail("could connect with invalid authentication token");
+    } catch (error) {
+      t.assert(error instanceof HolochainError);
+      assert(error instanceof HolochainError);
+      t.equal(error.name, "InvalidTokenError", "expected InvalidTokenError");
+    }
```
ThetaSinner, line 443: If at this point you call it again, I'd expect to get a "token is required", rather than "invalid token error"?

src/api/app/websocket.ts
jost-s: It is, and it's similar to the zome call signing credentials. They're equally stored in JavaScript memory for convenience, which compromises security to a degree.

test/e2e/index.ts
jost-s: This only calls once. What are you referring to by "you call it again"?

ThetaSinner: I see now this is the wrong test to ask this question. I'm looking for a test where we
- connect with a valid token that is single use,
- lose the connection then try to make a zome call, but now have an invalid token, 
- so we get a invalid token error
- then try to make another zome call and we should get a "no token" error or similar because the reconnect can't be attempted

ThetaSinner: Is that what the test below with the comment "Websocket reconnection has failed and subsequent calls should just return" is doing?

jost-s: Ah, yes, exactly. The subsequent call is attempted after not handling the invalid token error. In a real app, the invalid token error must be handled and a new token be requested before attempting another call.

src/api/client.ts
jost-s: Yes, absolutely! d3d8412bdde5aafd1c9074ca6b915bd3bd93a58d

test/e2e/index.ts
ThetaSinner: Sweet, okay. I'm caught up now and my original concern with this change has been addressed. Thank you :)

ThetaSinner (approved):

PR #283, opened 2024-10-29
Backport #279 to main-0.4

matthme:


.github/workflows/bundling-test.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - "main"
+      - "main-0.4"
```
jost-s: This is only for pushes, and generally pushing to main is a rare exception.

.github/workflows/integration-test.yml
```diff
@@ -4,8 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - "develop"
-      - "main"
+      - "main-0.4"
```
jost-s: Same here. We don't need `develop` here anymore, just `main`.

jost-s (changes requested):
LGTM, just the workflow changes
and please update the changelog

.github/workflows/bundling-test.yml
matthme: Oh, isn't it for pull requests too? I just wanted the tests to run on this PR (which is to main-0.4) to make sure I didn't mess up with my backporting.

matthme:
Okay, requested changes should be addressed.

.github/workflows/bundling-test.yml
jost-s: There's nothing after `pull_request:`, which means it's not restricted. `push:` is limited to branches, branches that are called `main`. So it should run on all PRs anyway.

CHANGELOG.md
```diff
@@ -5,12 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
 ### Added
+- Bring back a websocket reconnection automation for Admin and App websockets. When either of them is closed and a new request made, it will attempt to reconnect using the same app authentication token that was used to initially authenticate the websocket. A specific `InvalidTokenError` is returned if that fails.
 ### Changed
 - Update to Holochain 0.4.0-rc.0
 ### Fixed
 ### Removed
 
-## 2024-09-30: v0.18.0-dev.12
+## 2024-09-30: v0.18.0-dev.13
```
jost-s, line 14: :+1:

jost-s (approved):
Great!

PR #298, opened 2024-11-25
Update InstallAppRequest type to new format

matthme:
This PR contains the type changes required in order to adhere to https://github.com/holochain/holochain/pull/4416 and its follow-up PR https://github.com/holochain/holochain/pull/4488


jost-s (changes requested):
Please add a test or modify an existing one that uses roles_settings.

matthme:
I'm confused - why do the tests even pass? Are the tests not actually ran in CI?

jost-s:
The tests pass because `memproofs` is just ignored and `roles_settings` is optional.

matthme:
Test added.

matthme:
Hmm...actually the problem that `Uint8Array`s don't deserialize to `YamlProperties` is a problem. This means that you can't even pass in agent public keys there unless you base64 encode them or conver them to an `Array<number>`.

test/e2e/index.ts
```diff
@@ -562,6 +564,77 @@ test(
...
+        },
+        resources: {},
+      },
+      roles_settings: {
+        foo: {
+          type: "Provisioned",
+          membrane_proof: new Uint8Array(6),
+          modifiers: {
+            network_seed: "hello",
+            properties: yaml.dump({ progenitor: progenitorKey }),
```
matthme, line 613: Changed the test now to use a yaml string.
cc @ThetaSinner

matthme:
Okay never mind. I think this is totally appropriate and the app developer can be expected to take care of any required conversion for example using `js-yaml` (we just need to make sure that the js-client actually returns a useful error which is not currently the case https://github.com/holochain/holochain-client-js/issues/299).

package.json
```diff
@@ -48,13 +48,15 @@
     "isomorphic-ws": "^5.0.0",
     "js-base64": "^3.7.5",
     "js-sha512": "^0.9.0",
+    "js-yaml": "4.1.0",
```
jost-s: This can be a dev dependency too.

src/api/admin/types.ts
```diff
@@ -156,7 +156,43 @@ export type MembraneProof = Uint8Array;
...
+      type: "Provisioned";
+      membrane_proof?: MembraneProof;
+      modifiers?: DnaModifiersOpt;
+    };
+
+/**
+ *  @public
+ * Any value that is serializable to a Yaml value
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jost-s: Is this needed here? It should only produce a warning.
Actually `unknown` would be the idiomatic type.

test/e2e/common.ts
```diff
@@ -117,7 +117,7 @@ export const installAppAndDna = async (
     installed_app_id,
     agent_key: agent,
     path,
-    membrane_proofs: {},
+    roles_settings: {},
```
jost-s: `roles_settings` is optional
```suggestion
```

test/e2e/common.ts
```diff
@@ -159,7 +159,7 @@ export const createAppWsAndInstallApp = async (
...
```
jost-s:
```suggestion
```

test/e2e/index.ts
```diff
@@ -85,7 +87,7 @@ test(
       installed_app_id,
       agent_key,
       path: `${FIXTURE_PATH}/test.happ`,
-      membrane_proofs: {},
+      roles_settings: {},
```
jost-s: and so on further down
```suggestion
```

test/e2e/index.ts
```diff
@@ -1046,7 +1119,7 @@ test(
 
     const tag = "test_tag";
     const link: Link = await client.callZome({
-      cap_secret: null,
+      cap_secret: undefined,
```
jost-s:
```suggestion
```

jost-s (changes requested):
LGTM
a suggestion for the yaml properties type and some cleanup

matthme:
Thank you! Addressed all of these now.

Cargo.toml
```diff
@@ -7,4 +7,4 @@ resolver = "2"
 opt-level = "z"
 
 [workspace.dependencies]
-hdk = "0.5.0-dev.0"
+hdk = "0.5.0-dev.6"
```
jost-s: This is intentionally kept at `.0` as long as there are no breaking changes in the HDK.

test/e2e/common.ts
jost-s: I'm still here =)

jost-s (changes requested):
Just a few more bits

Cargo.toml
matthme: Alright!

matthme:
search and replace for the win.

test/e2e/common.ts
jost-s: And still here

test/e2e/common.ts
jost-s: Delete me

jost-s (changes requested):

test/e2e/common.ts
matthme: done.

jost-s (approved):

PR #300, opened 2024-11-28
Backport: New install app payload

matthme:


test/e2e/index.ts
```diff
@@ -1548,7 +1548,7 @@ test(
   })
 );
 
-test.only(
+test(
```
matthme, line 1549: I'm assuming that this was accidental...

jost-s:
I've fixed the tests now that DPKI is no longer present.

Cargo.toml
```diff
@@ -7,4 +7,4 @@ resolver = "2"
 opt-level = "z"
 
 [workspace.dependencies]
-hdk = "0.4.0-rc.0"
+hdk = "0.4.0-rc.1"
```
jost-s: please revert this

matthme:
Looks like tests passed - is it approvable now? =D

jost-s (changes requested):
One change please

Cargo.toml
matthme: It was actually necessary due to the __hc_preflight_request function I think.

matthme: Tests would fail otherwise

matthme: Hmm...I cannot reproduce it anymore. I reverted it. But I did get that error at some point and thought I had to bump the hdk for it to work...

jost-s (approved):

PR #308, opened 2024-12-17
Fix: dht op type helpers

matthme:
Fixes the DhtOp helper functions now that the DhtOp type changed to include warrants.


guillemcordoba (approved):

jost-s (approved):
LGTM

PR #314, opened 2024-12-18
chore(release): v0.18.0

matthme:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@holochain/client",
-  "version": "0.18.0-rc.2",
+  "version": "0.18.0",
```
ThetaSinner, line 3: package.lock should get updated when this changes?

matthme: indeed...updated now.

ThetaSinner (approved):

PR #318, opened 2025-01-03
Update cell provisioning types

### LINKED ISSUE #317: `CellProvisioning` doesn't match the enum in `holochain_types`
The `CellProvisioning` type in `src/api/admin/types.ts` doesn't closely match that of the corresponding Rust type, in either 0.4 or 0.5. There's a new `protected` flag for `UseExisting`; suggest using a tagged union rather than an interface. I'll follow up with a PR today.

pdaoust:
`CellProvisioning` broke compatibility at 0.4; there's now a `protected` flag for `UseExisting`. This adds that flag and also takes the opportunity to add `CloneOnly` which was previously missing.

I'd love some feedback on the TS idioms I've used here; I don't know how idiomatic or ergonomic they are.

Tests (including, but not only, the new ones I added) were failing locally due to API timeouts; hopefully that isn't masking any conductor errors (I believe I've fixed them all, but I'm not sure).

Closes #317 .


src/api/admin/types.ts
```diff
@@ -429,6 +427,12 @@ export type Location =
        * Get file from URL
        */
       url: string;
+    }
+  | {
```
pdaoust, line 433: This in particular feels weird to me -- an empty struct as a union variant. It's only there because `Location` gets intersected with `AppRoleDnaManifest`, because the corresponding stuff (that is, the optional `AppRoleDnaManifest::location` field's enum variant value) gets collapsed into `AppRoleDnaManifest` and we've got to find some similar way of modelling that.

pdaoust:
Tests were failing after the smoke cleared; converted to draft.

pdaoust:
Currently getting timeouts with the `installApp` method with an app with at least one role that uses `UseExisting` or `CloneOnly` provisioning strategies. Happens on 0.4 too. I'm not getting anything terribly interesting from the log.

pdaoust:
The `CloneOnly` strategy is blocked by an upstream bug holochain/holochain#4594 . The `UseExisting` test failure is a bit more rugose -- some sort of deserialisation error whereby it thinks I'm trying to pass a map of something as a cell ID, whereas I'm most certainly passing a `CellId`.

jost-s:
Coming back to this PR, I think it's too early to change provisioning types, as we haven't made progress on them in Holochain yet.

jost-s:
Closing this. Please re-open whenever needed.

PR #320, opened 2025-01-27
ci: Prepare CI for Pulumi management

ThetaSinner:


.github/workflows/test.yml
```diff
@@ -0,0 +1,174 @@
...
+      - name: Setup tmate session if a previous step failed
+        if: ${{ failure() }}
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 10
+
+  bundle:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
+        node: [ 20, 22 ]
```
ThetaSinner, line 82: Updated to LTS and previous LTS

.github/workflows/test.yml
```diff
@@ -0,0 +1,174 @@
...
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v4
```
ThetaSinner, line 88: Updated actions versions through these two jobs

.github/workflows/test.yml
```diff
@@ -0,0 +1,174 @@
...
+
+      - name: Install nix
+        uses: cachix/install-nix-action@v30
+        with:
+          install_url: https://releases.nixos.org/nix/nix-2.25.4/install
```
ThetaSinner, line 28: Bumped to latest Nix version

jost-s (approved):

matthme (commented):
I have two questions:
1. Is it correct that the bundling and the integration test are now running in sequence in a single workflow which will probably increase the time it takes for all tests to have ran? I'm not necessarily saying that this would be a problem but I'd like to understand it.
2. What is the context of this PR, i.e. why is it necessary for pulumi management as I infer from the PRs title?

matthme:
Wait, regarding 1. I think I mixed up jobs with steps. Jobs run in parallel I think.

ThetaSinner:
1. They run in parallel https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobs
2. I've done quite a few of these already but I think this should be possible to figure out from https://github.com/holochain/holochain-client-js/settings/rules/3424534 and https://github.com/holochain/hc-github-config/pull/6/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261
  It's standardizing the CI checks so that common rules can be enforced

matthme (approved):

matthme:
Okay thanks for the explanations :+1: . I don't have access to the repository rules (link 1) but I think I understand roughly what you mean now.

PR #322, opened 2025-02-17
Update enum serialization

matthme:
Updates types related to the changes of https://github.com/holochain/holochain/issues/4420


jost-s:
It doesn't build yet, there is a type error.

matthme:
It's supposed to build once holonix provides the new holochain verison with the serde flatten attribute removed (that's why it's a draft release still at the moment of writing).

matthme:
Oh sorry, I did not read your comment carefully. You're right, there's also a type error still.

src/api/admin/types.ts
```diff
@@ -120,22 +121,22 @@ export interface ClonedCell {
   enabled: boolean;
 }
 
-/**
- * @public
- */
-export enum CellType {
```
jost-s: I prefer to keep this. It prevents errors when mistyping the expected type.

src/api/app/websocket.ts
```diff
@@ -317,15 +322,15 @@ export class AppWebsocket implements AppClient {
         );
       }
       const cloneCell = appInfo.cell_info[baseRoleName].find(
-        (c) => CellType.Cloned in c && c[CellType.Cloned].clone_id === roleName
+        (c) => c.type === "cloned" && c.value.clone_id === roleName
       );
-      if (!cloneCell || !(CellType.Cloned in cloneCell)) {
+      if (!cloneCell) {
```
jost-s: I much prefer to repeat the condition here than a type assertion.
```suggestion
      if (!cloneCell || cloneCell.type !== CellType.Cloned) {
```

jost-s (changes requested):
Looking great, all the changes. Two changes requested.

src/api/admin/types.ts
matthme: Makes sense to me, yes. Added it back and updated the checks accordingly.

src/api/app/websocket.ts
matthme: Changed. It seems like a weak spot of typescript here that it doesn't know that an array filtered by type === "cloned" can only contain cells of type cloned, hence why I thought that check is redundant. But probably still worth to be explicit in checking.

jost-s: It is definitely lame.

jost-s (approved):

jost-s:
Ah damn, I forgot the changelog!

matthme:
no worries, I'll make a follow-up PR.

PR #329, opened 2025-03-10
Update AppBundleSource type and adjust tests accordingly

matthme:
### Summary

Updates the `AppBundleSource` type according to https://github.com/holochain/holochain/issues/4651 and adjusts the tests accordingly.

### TODO:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,3 +3,4 @@
 
 ### TODO:
 - [ ] CHANGELOG mentions all code changes.
+- [ ] docs have been updated (`npm run build:docs`)
--- a/test/e2e/index.ts
+++ b/test/e2e/index.ts
@@ -1,4 +1,4 @@
-import { decode } from "@msgpack/msgpack";
+import { decode, encode } from "@msgpack/msgpack";
 import assert from "node:assert/strict";
 import fs from "node:fs";
 import test from "tape";
@@ -34,6 +34,7 @@ import {
   ProvisionedCell,
   Duration,
   CellType,
+  AppBundle,
 } from "../../src";
 import {
   FIXTURE_PATH,
@@ -583,7 +584,6 @@ test(
 test(
   "can install app with roles_settings",
   withConductor(ADMIN_PORT, async (t) => {
-    const role_name = "foo";
     const installed_app_id = "app";
     const admin = await AdminWebsocket.connect({
       url: ADMIN_WS_URL,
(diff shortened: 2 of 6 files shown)


test/e2e/index.ts
```diff
@@ -603,28 +603,8 @@ test(
       installed_app_id,
       agent_key: agent,
       source: {
-        type: "bundle",
```
matthme, line 606: Installing the happ from a manifest was redundant here anyway so I changed it to installing from path.

test/e2e/index.ts
```diff
@@ -667,32 +647,40 @@ test(
...
+              strategy: CellProvisioningStrategy.Create,
+              deferred: false,
+            },
+            dna: {
+              bundled: "dna_1",
+              modifiers: { network_seed: "some_seed" },
+            },
+          },
+        ],
+        membrane_proofs_deferred: true,
```
matthme, line 669: This `membrane_proofs_deferred: true` setting here that deviates from the original manifest is the reason to bundle the happ with this special manifest explicitly here.

test/e2e/index.ts
```diff
@@ -769,106 +757,14 @@ test(
...
-    await admin.authorizeSigningCredentials(cell_id);
+    const appBundleBytes = fs.readFileSync(`${FIXTURE_PATH}/test.happ`);
 
-    const response = await client.callZome(zomeCallPayload, 30000);
-    t.equal(response, "foo", "zome call succeeds");
-  })
-);
-
-test(
-  "install app with app manifest and resource map",
```
matthme, line 832: We don't need this test anymore since the option to install from a manifest has been removed.

.github/pull_request_template.md
jost-s, line 1: :grinning:

test/e2e/index.ts
```diff
@@ -667,32 +647,40 @@ test(
...
+          },
+        ],
+        membrane_proofs_deferred: true,
+      },
+      resources: {
+        dna_1: zippedDnaBundle,
+      },
+    };
+
+    const zippedAppBundle = zlib.gzipSync(encode(appBundle));
```
jost-s, line 676: Why is this gzipped and not just encoded?

matthme: Because it expects the bytes of a .happ file which is gzipped. This is the error returned without it:
```
Test caught exception:  internal_error: Conductor returned an error while using a ConductorApi: AppBundleError(MrBundleError(StdIoError(Custom { kind: InvalidInput, error: "invalid gzip header" })))
```

matthme: https://github.com/holochain/holochain/blob/15cdd4bf429d0d0dcc02d03a18b9e732a35b066b/crates/mr_bundle/src/encoding.rs#L10 (that's the function used to encode the `AppBundle` to bytes in holochain).

jost-s: Right, so `bytes` is not exact.

jost-s (approved):

PR #335, opened 2025-04-06
Feat/await launcher env

mattyg:
### Summary
I have an issue without a perfect solution: https://github.com/holochain/android-service-runtime/issues/74 I believe this PR is least-worst option. 

This PR adds another magic variable `window.__HC_AWAIT_LAUNCHER_ENV__`. If this variable is defined, then the client will poll for `window.__HC_LAUNCHER_ENV__` until the specified timeout. 

This allows `window.__HC_LAUNCHER_ENV__` to be defined shortly after `.connect()` is called.

### TODO:
- [x] docs have been updated (`npm run build:docs`) (no changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## \[Unreleased\]
 
 ### Added
+- Add new magic configuration variable `window.__HC_AWAIT_LAUNCHER_ENV__`. If this variable is defined, then `AppWebsocket` and `AdminWebsocket`s' `connect` function will poll for `window.__HC_LAUNCHER_ENV__`, until a timeout is reached. This allows `window.__HC_LAUNCHER_ENV__` to be defined shortly *after* `connect` is called.
+
 ### Fixed
 ### Changed
 ### Removed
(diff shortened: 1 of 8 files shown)


matthme:
I haven't yet looked much into how the general logic is set up but one thing I was wondering is whether the port and token couldn't be passed along in the [main_window_builder()](https://github.com/holochain/android-service-runtime/blob/72c3431db02611a81b82b5a9a6017fed755970ad/crates/tauri-plugin-holochain-service-consumer/src/mobile.rs#L42-L47) so it wouldn't need to be fetched via IPC?
I haven't found where `main_window_builder()` is actually being invoked, is it not a process that has access to the admin websocket and could create the token and pass it along?

mattyg:
`main_window_builder` is invoked when the consumer tauri app is setup: https://github.com/holochain/android-service-runtime/pull/61/files#diff-1c0e486421c66efcf0589bf36286a630503da79e90470ab59fd33c4b5ed9e9d5R20

It doesn't use or create an admin websocket to avoid exposing it, but instead calls the conductor handle directly, so with that approach the calls do have to be over IPC.

But I think I could make the IPC calls from the rust side. I went down that path early on but ran into some issues I no longer remember and abandoned it. I'll look into that again now that I understand the stack much better. It would be much cleaner.

mattyg:
The refactoring worked out: https://github.com/holochain/android-service-runtime/pull/78 Thanks for the advice.

matthme:
sweet!

PR #341, opened 2025-04-17
fix: InstalledAppInfoStatus type

matthme:
### Summary
- fixes the `InstalledAppInfoStatus` type.

### TODO:


mattyg (commented):
I think `PausedAppReason` and `DisabledAppReason` need to be updated as well, right?

Also I see there is a type `DeactivationReason` that I can't find used anywhere - can it be removed?

matthme:
You catch my sloppiness in bunches. Thanks! I updated the other types as well, removed `DeactivationReason` and also renamed `InstalledAppInfoStatus` to `AppInfoStatus` as it's being called on the holochain side for a while now.

mattyg (approved):

PR #349, opened 2025-06-19
Add agent info to app ws

jost-s:
### Summary
Tests of examples in the README didn't return exit code of errors. A breaking change went unnoticed - something the tests were meant to prevent. No clue why the errors of the most important part of the bash script were not to exit the script. Fixed now.

### TODO:


doc-test-app-ws.sh
```diff
@@ -8,8 +8,6 @@ HC_ID=$!
 echo "HC_ID is $HC_ID"
 sleep 5
 
-set +e
```
ThetaSinner, line 11: Oh yes, that doesn't look wanted!

test/e2e/common.ts
```diff
@@ -1,23 +1,86 @@
...
-import { AppWebsocket, CoordinatorBundle } from "../../src";
-import fs from "fs";
 
 export const FIXTURE_PATH = "./test/e2e/fixture";
 
+const BOOTSTRAP_SERVER_STARTUP_STRING = "#kitsune2_bootstrap_srv#listening#";
 const LAIR_PASSPHRASE = "passphrase";
 
-export const launch = async (port: number) => {
+export const runLocalServices = async () => {
```
ThetaSinner, line 20: You know what'd be really neat if there was endless time? Embedding these tools as native libraries and just exposing a JS interface for them, then including them in tools like this should be simpler

ThetaSinner (approved):
Nice! I really appreciate that we're starting to get tooling upgraded for 0.6 before we make so many changes that it becomes hard to know where to start, there's already quite a bit of change clearly!

test/e2e/common.ts
jost-s: Yes! Or at least share this wrapper between Tryorama and the JS client.

PR #352, opened 2025-06-26
Backport 0.5: App agent info request

matthme:
### Summary

Backport of #349 

### TODO:


src/api/app/types.ts
```diff
@@ -82,6 +82,11 @@ export type CallZomeResponse = CallZomeResponseGeneric<any>;
  */
 export type AppInfoResponse = AppInfo | null;
 
+/**
+ * @public
+ */
+export type AppAgentInfoRequest = { dna_hashes: DnaHash[] | null };
```
matthme, line 88: Since the admin call still takes `CellId` we need to call the app request type differently and can't reuse the `AgentInfoRequest` type from the admin websocket.

ThetaSinner (approved):
Looks like a tidy backport to me, and I like the decision to just split the types where there's a difference between app and admin calls

PR #353, opened 2025-06-26
feat(main-0.5): Add agent meta info request to app and admin websocket

matthme:
### Summary

Add agent meta info request to app and admin websocket 

### TODO:


src/api/admin/types.ts
```diff
@@ -654,6 +654,27 @@ export type AddAgentInfoRequest = { agent_infos: Array<string> };
...
 
+/**
+ * @public
+ */
+export type AgentMetaInfoRequest = { url: string; dna_hashes?: DnaHash[] };
+
+/**
+ * @public
+ */
+export type AgentMetaInfoReqsponse = Record<
```
jost-s:
```suggestion
export type AgentMetaInfoResponse = Record<
```

src/api/admin/types.ts
```diff
@@ -654,6 +654,27 @@ export type AddAgentInfoRequest = { agent_infos: Array<string> };
...
+  DnaHashB64,
+  Record<string, AgentMetaInfo>
+>;
+
+/**
+ * @public
+ */
+export type AgentMetaInfo = {
```
jost-s:
```suggestion
export interface AgentMetaInfo = {
```

src/api/admin/types.ts
```diff
@@ -654,6 +654,27 @@ export type AddAgentInfoRequest = { agent_infos: Array<string> };
  */
 export type AddAgentInfoResponse = unknown;
 
+/**
+ * @public
+ */
+export type AgentMetaInfoRequest = { url: string; dna_hashes?: DnaHash[] };
```
jost-s: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

"use interfaces until you need to use a type"
```suggestion
export interface AgentMetaInfoRequest = { url: string; dna_hashes?: DnaHash[] };
```

test/e2e/index.ts
```diff
@@ -982,6 +985,100 @@ test("can inject agents", async (t) => {
...
+        provenance: cell_id2[1],
+        zome_name: TEST_ZOME_NAME,
+        fn_name: "get_an_entry",
+        payload: acionHash,
+      }),
+    null,
+    "agent 2 wasn't able to get the entry of agent 1",
+    200,
+    20_000
+  );
```
ThetaSinner, line 1048: `dhtSync` here?

ThetaSinner (dismissed):
Nice!

test/e2e/index.ts
matthme: `dhtSync` is in tryorama so I don't think I can use it here (tryorama depends on the js-client so importing tryorama would be a circular dependency).

test/e2e/common.ts
```diff
@@ -298,3 +298,25 @@ export async function makeCoordinatorZomeBundle(): Promise<CoordinatorBundle> {
     },
   };
 }
+
+export async function retryUntilTimeout<T>(
```
jost-s, line 302: It's a matter of preference too, but I suggest to match what is done in Holochain and kitsune2, where the callback returns a `Promise<boolean>`, and the loop runs until the boolean is true.

jost-s (changes requested):
Looks good. I request to fix the typo and change some typescript exports. The `retry` fn may stay the way it is.

src/api/admin/types.ts
matthme: https://github.com/holochain/holochain-client-js/pull/353/commits/50ca7f8bba6d834d514fc90a23d99d4b873fa2b6

src/api/admin/types.ts
matthme: https://github.com/holochain/holochain-client-js/pull/353/commits/50ca7f8bba6d834d514fc90a23d99d4b873fa2b6

src/api/admin/types.ts
matthme: https://github.com/holochain/holochain-client-js/pull/353/commits/50ca7f8bba6d834d514fc90a23d99d4b873fa2b6

test/e2e/common.ts
matthme: Hmm...I thought it might in some cases be actually useful to return a value. In case you fetch a Record and then want to use the Record's content in the test as well for example. Might be thinking too far though because in this case I'm not actually using the content. I just thought it's more flexible to allow arbitrary return types.

test/e2e/common.ts
```diff
@@ -298,3 +298,25 @@ export async function makeCoordinatorZomeBundle(): Promise<CoordinatorBundle> {
...
+  whileValue: any,
+  timeoutMsg: string,
+  intervalMs: number,
+  timeoutMs: number
+): Promise<T> {
+  const startTime = Date.now();
+  let value = await cb();
+  while (value === whileValue) {
+    const currentTime = Date.now();
+    if (Math.floor(currentTime - startTime) >= timeoutMs)
```
jost-s: `Date.now()` always returns integers, so flooring doesn't do anything.
```suggestion
    if (currentTime - startTime >= timeoutMs)
```

test/e2e/common.ts
jost-s: Alright, fine by me.

test/e2e/common.ts
matthme: Oh, good point. I blind-copied that over from tryorama.

matthme: https://github.com/holochain/holochain-client-js/pull/353/commits/06a29c32ada14f1fc6a2d5b0502334dde88f43f6

jost-s (approved):

PR #354, opened 2025-06-27
fix: make WsClientOptions undefined by default due to lacking browser support

matthme:
### Summary

Providing an `options` argument to the `Websocket` is apparently not supported in the browser, see this issue https://github.com/heineiuo/isomorphic-ws/issues/15 and the correspondingly [updated docs](https://github.com/heineiuo/isomorphic-ws/pull/17/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R20) that state

> no support for the contructor  [`options`](https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketaddress-protocols-options) argument in browsers

As a result the `reconnectWebsocket()` method fails [here](https://github.com/holochain/holochain-client-js/blob/bc746b843a0111e8094c617480dde3e931389d26/src/api/client.ts#L297) when the js-client is being used in the browser because an empty object `{}` is being passed along by default.

This PR fixes this by passing along `undefined` by default.

### TODO:


jost-s:
Okay, in the browser the origin is set automatically anyway, right? That's all that those options are used for.

matthme:
> Okay, in the browser the origin is set automatically anyway, right? That's all that those options are used for.

Exactly. In the browser there's no way to override the origin anyway.

jost-s (approved):

PR #361, opened 2025-07-21
fix: Agent meta info declares the wrong response type

ThetaSinner:
### TODO:
- [ ] CHANGELOG mentions all code changes.
- [ ] docs have been updated (`npm run build:docs`)

### DIFF
--- a/src/api/app/websocket.ts
+++ b/src/api/app/websocket.ts
@@ -13,11 +13,9 @@ import {
   AgentInfoRequest,
   AgentInfoResponse,
   AgentMetaInfoRequest,
+  AgentMetaInfoResponse,
   AppInfo,
   CellType,
-  // Required to TSDoc generation.
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  ClonedCell,
   DumpNetworkMetricsRequest,
   DumpNetworkMetricsResponse,
   DumpNetworkStatsRequest,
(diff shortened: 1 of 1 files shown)


matthme (changes requested):
There seem to be quite a few formatting changes. Could you run `npm run format` to undo them? Maybe we should add that command to the `build` npm script as well...

veeso (approved):

src/api/app/websocket.ts
```diff
@@ -13,11 +13,9 @@ import {
   AgentInfoRequest,
   AgentInfoResponse,
   AgentMetaInfoRequest,
+  AgentMetaInfoResponse,
   AppInfo,
   CellType,
-  // Required to TSDoc generation.
```
matthme, line 18: It seems that the removal of this doesn't affect the result of `npm run build:docs` so it's probably fine to remove it now? I wonder what it was required for exactly in the first place though given that it has an explicit comment that it's required despite being unused.

matthme (approved):

src/api/app/websocket.ts
ThetaSinner: Yeah I'm not sure about this. I didn't remove it manually so it must have been my IDE detecting it unused. I'm hoping the "this is needed" part was just about the TypeScript directive on the import and not the import itself.

PR #367, opened 2025-08-04
chore: remove registerDna admin call

matthme:
### Summary

Removes the registerDna admin call to follow https://github.com/holochain/holochain/pull/5175

### TODO:

## Summary by CodeRabbit

* **Removed**
  * The ability to register a DNA via the admin API has been removed.
  * Documentation and references for DNA registration and related request/response types have been deleted.
  * End-to-end tests related to DNA registration have been removed or updated.

* **Documentation**
  * Updated documentation to reflect the removal of DNA registration functionality from the admin API.


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Removed
 
+- **BREAKING** Removed `RegisterDna` admin call ([#366](https://github.com/holochain/holochain-client-js/pull/366))
```
cdunster:
```suggestion
- **BREAKING** Removed `RegisterDna` admin call ([#367](https://github.com/holochain/holochain-client-js/pull/367))
```

cdunster (changes requested):
I think that the changelog entry points to the wrong PR

CHANGELOG.md
matthme: Thanks! Addressed.

cdunster (approved):
List DNAs is still a thing, and the docs say that it "lists all registered DNAs". I'm not sure if that also needs removing or if the docs should be updated?

matthme:
> List DNAs is still a thing, and the docs say that it "lists all registered DNAs". I'm not sure if that also needs removing or if the docs should be updated?

Yes I saw that too and was unsure whether I should rename the doc string there. I think removing the call altogether would be premature since it hasn't been removed in holochain yet (and there may be reasons to keep it).
And I didn't know a better word for "registered". It's also not quite "instantiated" because it returns the dna hashes associated to ribosomes in the ribosome store and that doesn't necessarily need to correspond to instantiated cells. Although I'm not 100% sure about that, now that the registerDna call has been removed maybe there's no way to have a ribosome in a store for a dna hash that has no instantiated cell.

ThetaSinner (approved):

PR #368, opened 2025-08-04
fix: expires_at field in PeerMetaInfo type can be undefined

matthme:
### Summary

As mentioned in [this PR comment](https://github.com/holochain/holochain-client-js/pull/364#discussion_r2248170088) and fixed in holochain with https://github.com/holochain/holochain/pull/5183, the `PeerMetaInfo` type's `expires_at` field can be undefined.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified that the `expires_at` field in `PeerMetaInfo` can now be null, improving accuracy in documentation and type definitions.

### DIFF
--- a/src/api/admin/types.ts
+++ b/src/api/admin/types.ts
@@ -646,7 +646,7 @@ export type PeerMetaInfoResponse = Record<
  */
 export interface PeerMetaInfo {
   meta_value: string;
-  expires_at: Timestamp;
+  expires_at: Timestamp | null;
 }
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed `expires_at` field in `PeerMetaInfo` type to allow it being `undefined` ([#368](https://github.com/holochain/holochain-client-js/pull/368))
 - Fixed types for `list_apps` to match the new API
 
 ### Changed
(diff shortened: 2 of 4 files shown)


src/api/admin/types.ts
```diff
@@ -646,7 +646,7 @@ export type PeerMetaInfoResponse = Record<
  */
 export interface PeerMetaInfo {
   meta_value: string;
-  expires_at: Timestamp;
+  expires_at: Timestamp | undefined;
```
veeso:
```suggestion
  expires_at?: Timestamp;
```

veeso (changes requested):

src/api/admin/types.ts
matthme: I had it like this first but then changed it to `| undefined` because I thought it's not an input parameter type but a type you get in a response so I thought the question mark is less appropriate in such a case. And I saw other usages of `| undefined` for output types as well in the code base. But I changed it to `?` now, I'm not strongly attached.

matthme: In fact, the correct type would probably be `Timestamp | null` since I think a `None` value [gets deserialized](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a4ecc54186a67b2c4237d88dcf2abeaa) as `null`. The question mark would indicate that a field does not need to be there at all (which is appropriate for request payloads where a missing field, if expected on the Rust side, automatically gets serialized to `None`) but I think in this case the field will always be there but with an explicit `null` value.

ThetaSinner (approved):

veeso (approved):

PR #370, opened 2025-08-15
fix: update hc sandbox command used by tests to changes introduced in holochain 0.6.0-dev.16

matthme:
### Summary
The `-e` option in `hc sandbox run` has been removed with https://github.com/holochain/holochain/pull/5173. This PR updates the command accordingly.

### TODO:

## Summary by CodeRabbit

* **Documentation**
  * Changelog updated to note internal test sandbox alignment with Holochain 0.6.0-dev.16; documentation-only, no code/API/runtime changes.

* **Tests**
  * End-to-end tests updated to select the target conductor by index for sandbox runs.
  * Added error handling when the conductor index cannot be determined.

### DIFF
--- a/test/e2e/common.ts
+++ b/test/e2e/common.ts
@@ -102,10 +102,22 @@ export const launch = async (
   });
   await createConductorPromise;
 
+  // Find out the conductor index of the created conductor
+  const dotHcFileContentLines = fs
+    .readFileSync(".hc", "utf-8")
+    .split(/\r?\n/)
+    .map((l) => l.trim());
+  const conductorIndex = dotHcFileContentLines.findIndex(
+    (line) => line === conductorDir
+  );
+
+  if (conductorIndex === -1)
+    throw new Error("Failed to determine index of recently started conductor.");
+
   // start sandbox conductor
   const runConductorProcess = spawn(
     "hc",
-    ["sandbox", "--piped", `-f=${port}`, "run", "-e", conductorDir],
+    ["sandbox", "--piped", `-f=${port}`, "run", `${conductorIndex}`],
     {
       detached: true, // create a process group; without this option, killing
       // the process doesn't kill the conductor
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Updated hc sandbox command used in internal tests according to changes introduced with holochain 0.6.0-dev.16 ([#370](https://github.com/holochain/holochain-client-js/pull/370))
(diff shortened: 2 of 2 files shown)


test/e2e/common.ts
```diff
@@ -85,17 +85,10 @@ export const launch = async (
   createConductorProcess.stdin.write(LAIR_PASSPHRASE);
   createConductorProcess.stdin.end();
 
-  let conductorDir = "";
   const createConductorPromise = new Promise<void>((resolve) => {
     createConductorProcess.stderr.on("data", (data) => {
       console.error("[hc sandbox] ERROR: ", data.toString());
     });
-    createConductorProcess.stdout.on("data", (data) => {
```
matthme: Not required anymore since we can run an existing sandbox by index now rather than by specifying the conductor directory.

ThetaSinner (dismissed):
Nice, looks good

.github/workflows/test.yml
```diff
@@ -43,10 +43,10 @@ jobs:
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
 
       - name: Build Nix packages
-        run: nix develop -c $SHELL -c "holochain --version"
+        run: nix develop -c $SHELL -c "holochain --version || exit 1"
```
cdunster: Why is the `|| exit 1` needed? Is there something wrong with the status codes?

matthme: I'm trying to get failed tests bubble up to have the workflow fail. But the `|| exit 1` doesn't seem to help either. Currently, if a test fails then the workflow just runs until it times out after 6 hours instead of failing.

test/e2e/common.ts
```diff
@@ -102,10 +102,19 @@ export const launch = async (
   });
   await createConductorPromise;
 
+  // Find out the conductor index of the created conductor
```
matthme, line 105: Turns out I need to actually figure out the correct conductor index or else tests that spawn two conductors will end up using the same conductor.

test/e2e/common.ts
```diff
@@ -102,10 +102,19 @@ export const launch = async (
...
+  const dotHcFileContentLines = fs.readFileSync(".hc", "utf-8").split("\n");
+  const conductorIndex = dotHcFileContentLines.findIndex(
+    (line) => line === conductorDir
+  );
+
+  if (conductorIndex === undefined)
+    throw new Error("Failed to determine index of recently strated conductor.");
+
```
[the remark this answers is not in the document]
matthme: Good points mostly, thank you. The Boolean filter in the end is unnecessary I think.

.github/workflows/test.yml
matthme: Removed it again. The problem seems to be rather that one of the processes (holochain and/or lair) keeps running despite the failure.

ddd-mtl (approved):

PR #414, opened 2026-05-15
fix: prevent WsClient wedge after transient reconnect (#412)

### LINKED ISSUE #412: Bug: `WsClient` permanently wedges on transient reconnect failure
# Bug: `WsClient` permanently wedges on transient reconnect failure

## TL;DR

There are transient connection failures (laptop close/open) that turn into permanent re-connection failures which break the client's auto-reconnect logic because of the way `invalidTokenCloseListener` is a 10ms heuristic for authentication failure, (which isn't always what the failure was) and clears the authToken (which may actually still be valid) and then won't try to reconnect, which would actually work.

## Affected

* Branch: `main-0.6` (verified at `0.20.4-rc.0`)
* Branch: `main` (`0.21.0-dev.2`) — `git diff main-0.6..main -- src/api/client.ts` is empty, same code, same bug.
* Not present on `develop` (`0.17.0-dev.9`) — that branch predates the auto-reconnect feature entirely.
* Introduced by [#279](<https://github.com/holochain/holochain-client-js/pull/279>) (commit `d6b8893`),<br>backported as [#283](<https://github.com/holochain/holochain-client-js/pull/283>) (commit `81effb8`).

## Symptom

After an initially healthy `AppWebsocket` connection experiences a single<br>transient reconnect failure (e.g. brief network blip, conductor momentarily<br>unreachable, server-initiated close racing with a request), the client enters<br>a state where every subsequent request rejects with:

```
HolochainError { name: 'WebsocketClosedError', message: 'Websocket is not open' }
```

— forever, with no recovery path. The user must reload to escape.

In moss this manifests as console spam from any code that polls through a<br>captured `AppWebsocket` reference (the debugging panel's `dumpNetworkMetrics`<br>loop, profile/peer-status polling via `GroupStore`, etc.).

## Root Cause

`WsClient.exchange()` ([src/api/client.ts:161-187](<src/api/client.ts#L161-L187>)) has three branches:

```typescript
if (this.socket.readyState === this.socket.OPEN) {
  // (A) Send normally
} else if (this.url && this.authenticationToken) {
  // (B) Reconnect via reconnectWebsocket(url, token), then send
} else {
  // (C) Reject with WebsocketClosedError
}
```

Branch (B) is gated on `this.authenticationToken`. The recovery path can only<br>fire while the token field is set.

`reconnectWebsocket` ([src/api/client.ts:295-345](<src/api/client.ts#L295-L345>)) clears that field on **any****<br>****non-success outcome of the new connection attempt**:

* `socket.addEventListener("error", ...)` — sets `this.authenticationToken = undefined`.
* `invalidTokenCloseListener` (any `close` event within \~10ms of opening) —<br>sets `this.authenticationToken = undefined`.

Both clears are eager. The 10ms heuristic is meant to detect "conductor<br>rejected our token," but it also fires for any transient close (server kicked<br>the connection, race during conductor boot, intermittent network). The<br>`"error"` clear fires on connection refusal, which is often transient too.

Once the field is cleared, `exchange()` falls into branch (C) on every<br>subsequent request and there is no code path that ever re-populates the field<br>or re-attempts authentication. The client is permanently stuck.

The token in the conductor's in-memory store is typically still valid (e.g.<br>moss issues with `single_use: false, expiry_seconds: 0`), and the consumer's<br>own cached copy is fine — only the `WsClient`'s private copy has been nulled,<br>and that's the one `exchange()` checks.

## Reproduction

See test below or in any browser/Electron context with a live `AppWebsocket`:

```js
const wsClient = appWs.client;
const goodUrl = wsClient.url;

// Sabotage: point the client at a closed port, then close the live socket.
// The next request enters exchange() → not OPEN → reconnect to bad URL →
// "error" event fires → authenticationToken cleared → reject with
// ConnectionError. From this moment on the client is wedged.
wsClient.url = new URL('ws://127.0.0.1:1');
wsClient.socket.close();

try { await appWs.dumpNetworkStats(); } catch (e) { console.log(e.name); }
// → 'ConnectionError'

// Restore URL — but the token is already cleared.
wsClient.url = goodUrl;

try { await appWs.dumpNetworkStats(); } catch (e) { console.log(e.name); }
// → 'WebsocketClosedError'  (forever, no recovery)
```

Equivalent real-world trigger: kill and respawn the conductor while the client<br>is mid-request, or any TCP-level interruption that races with a<br>`reconnectWebsocket` attempt.

## Why It Matters

* Affects every long-running app that uses `@holochain/client` against an<br>app interface.
* Single transient hiccup → permanent client breakage.
* Symptom is "the UI silently stops working" or "console spam," neither of<br>which has any user-actionable diagnostic.
* Consumer-side workarounds exist (sticky-token monkey-patch, periodic<br>re-creation of the `AppWebsocket`) but are fragile and don't help long-lived<br>references that have already been distributed to subsystems.

## Notes for the Fix Author

* The reproduction snippet above is verifiable in any DevTools console<br>attached to a working consumer (Moss, etc.).
* The existing tests for the reconnect feature live under `test/e2e/index.ts`<br>(added in [#283](<https://github.com/holochain/holochain-client-js/pull/283>));<br>a regression test should cover "reconnect attempt fails, then succeeds, then<br>the same client makes a successful request."

Here is a test that can be added to that file that shows the issue:

```typescript
test(
  "BUG: client wedges permanently after a transient reconnect failure",
  withConductor(ADMIN_PORT, async (t) => {
    // single_use=false, expiry=0 -> token stays valid for the whole session.
    // Rules out "token actually expired" as a cause of failures below.
    const { cell_id, client, admin } = await installAppAndDna(
      ADMIN_PORT,
      false,
      0,
    );
    await admin.authorizeSigningCredentials(cell_id);

    const callParams = {
      cell_id,
      zome_name: TEST_ZOME_NAME,
      fn_name: "bar",
      provenance: cell_id[1],
      payload: null,
    };

    // Sanity: a healthy call works.
    await client.callZome(callParams);
    t.pass("baseline zome call succeeded");

    // Induce a transient reconnect failure: point at a closed port and close
    // the live socket. The next request enters exchange() -> not OPEN ->
    // reconnect to bad URL -> error/early-close fires -> authenticationToken
    // is cleared inside reconnectWebsocket.
    const goodUrl = client.client.url!;
    const badUrl = new URL("ws://127.0.0.1:1");
    client.client.url = badUrl;
    await client.client.close();

    let transientErrName: string | undefined;
    try {
      await client.callZome(callParams);
      t.fail("call against sabotaged URL should have failed");
    } catch (error) {
      assert(error instanceof HolochainError);
      transientErrName = error.name;
      t.pass(
        `transient reconnect failure observed (error name: ${error.name})`,
      );
    }
    t.ok(transientErrName, "transient failure surfaced as HolochainError");

    // Restore the good URL. Conductor is up, token is still valid.
    client.client.url = goodUrl;

    // BUG: client is now permanently wedged. Subsequent calls reject with
    // WebsocketClosedError because authenticationToken was nulled and there
    // is no path that re-populates it. This test PASSES today because the
    // wedge is real; once fixed, flip to expect success.
    try {
      await client.callZome(callParams);
      t.fail(
        "BUG NOT REPRODUCED: client recovered after URL restore - either " +
          "the bug is fixed or the repro path no longer triggers it",
      );
    } catch (error) {
      assert(error instanceof HolochainError);
      t.equal(
        error.name,
        "WebsocketClosedError",
        "BUG CONFIRMED: client wedged with WebsocketClosedError despite " +
          "valid token, healthy conductor, and correct URL",
      );
    }

    // Confirm the wedge persists (no self-recovery).
    try {
      await client.callZome(callParams);
      t.fail("second post-recovery call unexpectedly succeeded");
    } catch (error) {
      assert(error instanceof HolochainError);
      t.equal(
        error.name,
        "WebsocketClosedError",
        "wedge persists across multiple calls",
      );
    }
  }),
);
```

**AC:**

* Add reproduction as regression test
* Don't unset token in client when a general error occurs.

veeso:
### Summary

closes #412

- Retain auth token on generic reconnect "error" event; only clear it
  when conductor closes immediately after `authenticate` (real
  InvalidTokenError signal).
- Dedupe concurrent reconnect attempts via shared `reconnectPromise`
  so parallel `exchange()` callers share one new socket.
- Document `request()` error surface (ConnectionError /
  InvalidTokenError / WebsocketClosedError).
- Add e2e regressions: wedge recovery + concurrent reconnect dedupe.
- Switch tsconfig to NodeNext module/moduleResolution; fix resulting
  missing `.js` extension in `src/hdk/validation-receipts.ts`. This was necessary because our tsconfig was going to raise error on next ts version

### TODO:
- [ ] CHANGELOG mentions all code changes.
- [ ] docs have been updated (`npm run build:docs`)

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -42,6 +42,7 @@ export class WsClient extends Emittery {
   private pendingRequests: Record<number, HolochainRequest>;
   private index: number;
   private authenticationToken: AppAuthenticationToken | undefined;
+  private reconnectPromise: Promise<void> | undefined;
 
   constructor(socket: IsoWebSocket, url?: URL, options?: WsClientOptions) {
     super();
@@ -151,8 +152,21 @@ export class WsClient extends Emittery {
   /**
    * Send requests to the connected websocket.
    *
+   * If the underlying socket is closed when this method is called, the
+   * client transparently reconnects and re-authenticates using the cached
+   * token. Transient reconnect errors are surfaced as a `ConnectionError`
+   * and the cached token is retained, so a subsequent call can retry the
+   * reconnect.
+   *
+   * If the conductor rejects the cached token during the reconnect
+   * (signalled by an immediate close after the `authenticate` handshake),
+   * the cached token is cleared and the call rejects with an
+   * `InvalidTokenError`. The consumer must rebuild the `AppWebsocket`
+   * with a fresh token; further calls on this client will reject with
+   * `WebsocketClosedError`.
+   *
    * @param request - The request to send over the websocket.
-   * @returns
+   * @returns The decoded response payload.
    */
   async request<Response>(request: unknown): Promise<Response> {
     return this.exchange(request, this.sendMessage.bind(this));
@@ -172,7 +186,18 @@ export class WsClient extends Emittery {
       });
       return promise as Promise<Response>;
     } else if (this.url && this.authenticationToken) {
-      await this.reconnectWebsocket(this.url, this.authenticationToken);
+      // Dedupe concurrent reconnect attempts. The first caller into this
+      // branch starts a single reconnect; further callers await the same
+      // promise so we never create multiple sockets in parallel.
+      if (!this.reconnectPromise) {
+        this.reconnectPromise = this.reconnectWebsocket(
+          this.url,
+          this.authenticationToken,
+        ).finally(() => {
+          this.reconnectPromise = undefined;
+        });
+      }
+      await this.reconnectPromise;
       this.registerMessageListener(this.socket);
       this.registerCloseListener(this.socket);
       const promise = new Promise((resolve, reject) =>
@@ -295,11 +320,16 @@ export class WsClient extends Emittery {
   private async reconnectWebsocket(url: URL, token: AppAuthenticationToken) {
     return new Promise<void>((resolve, reject) => {
       this.socket = new IsoWebSocket(url, this.options);
+
+      // Track whether the "open" event has fired. The invalidTokenCloseListener
+      // must only clear the token when the connection did open and the conductor
+      // then closed it quickly.
+      let openFired = false;
+
       // This error event never occurs in tests. Could be removed?
       this.socket.addEventListener(
         "error",
         (errorEvent) => {
-          this.authenticationToken = undefined;
           reject(
             new HolochainError(
               "ConnectionError",
@@ -313,13 +343,22 @@ export class WsClient extends Emittery {
       const invalidTokenCloseListener = (
         closeEvent: IsoWebSocket.CloseEvent,
       ) => {
-        this.authenticationToken = undefined;
-        reject(
-          new HolochainError(
-            "InvalidTokenError",
-            `could not connect to ${this.url} due to an invalid app authentication token - close code ${closeEvent.code}`,
-          ),
-        );
+        if (openFired) {
+          this.authenticationToken = undefined;
+          reject(
+            new HolochainError(
+              "InvalidTokenError",
+              `could not connect to ${this.url} due to an invalid app authentication token - close code ${closeEvent.code}`,
+            ),
+          );
+        } else {
+          reject(
+            new HolochainError(
+              "ConnectionError",
+              `could not connect to Holochain Conductor API at ${url} - close code ${closeEvent.code}`,
+            ),
+          );
+        }
       };
       this.socket.addEventListener("close", invalidTokenCloseListener, {
         once: true,
@@ -328,6 +367,7 @@ export class WsClient extends Emittery {
       this.socket.addEventListener(
         "open",
         async () => {
+          openFired = true;
           const encodedMsg = encode({
             type: "authenticate",
             data: encode({ token }),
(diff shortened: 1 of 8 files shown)


tsconfig.json
```diff
@@ -1,16 +1,14 @@
 {
   "compilerOptions": {
-    "module": "ES2022",
+    "module": "NodeNext",
     "target": "ES2022",
-    "moduleResolution": "Node",
```
veeso, line 5: Deprecated, would have raised error soon

src/api/client.ts
```diff
@@ -172,7 +186,18 @@ export class WsClient extends Emittery {
...
+      // promise so we never create multiple sockets in parallel.
+      if (!this.reconnectPromise) {
+        this.reconnectPromise = this.reconnectWebsocket(
+          this.url,
+          this.authenticationToken,
+        ).finally(() => {
+          this.reconnectPromise = undefined;
+        });
+      }
+      await this.reconnectPromise;
```
ThetaSinner, line 200: Nice

ThetaSinner (approved):
Looks great, covers everything I'd want it to

PR #425, opened 2026-06-03
feat: route the App API over Tauri IPC for in-process plugin webviews

zippy:
## Summary

Adds a **Tauri IPC transport** for the App API so a UI running inside an in-process Tauri plugin webview can reach a same-process conductor without a loopback websocket. The websocket path is untouched and remains the transport for the separated client + service deployment and any non-Tauri consumer.

When the plugin injects its `__HC_TAURI_HOLOCHAIN__` marker on `window`, `AppWebsocket.connect()` transparently builds a `TauriAppTransport` instead of a `WsClient` — no consumer code change. `callZome`, `appInfo`, signals, clone cells, etc. all work unchanged because the transport is a drop-in for the slice of `WsClient` that `AppWebsocket` uses.

## Design

Same bytes, different pipe: the unit of transfer is the msgpack-encoded tagged `{ type, value }` payload — identical to what the websocket carries — moved through the `plugin:holochain|app_request` command and decoded with the same `Uint8Array`→Base64 map-key converter `WsClient` uses, so HoloHash-keyed maps deserialize identically. No new DTOs, no JSON, no parallel type definitions.

- `src/environments/tauri.ts` — `isTauriHolochain()` detector + injected-env accessor (mirrors `launcher.ts`).
- `src/api/app/tauri-transport.ts` — `TauriAppTransport`: `request()` over IPC + signal bridge re-emitted on Emittery.
- `src/api/app/websocket.ts` — `connect()` branches to the Tauri transport ahead of the launcher-env websocket rewrite; the transport field is relaxed to a minimal `AppClientTransport` interface that `WsClient` already satisfies.
- `test/tauri.ts` — 22 pure unit tests (mocked `invoke`, no conductor) locking in the encode/dispatch/decode contract, signal delivery, and host-signed zome calls. Wired into CI.

## Trust model

The websocket path is scoped by a per-app auth token. The Tauri path replaces that with: (a) the conductor derives the app id from the **calling window's label**, never from a JS-supplied argument, and (b) Tauri's capability/permission system gates the `app_request` command to the intended windows. Zome calls are still signed by the in-process keystore via the host signer — no signing key is exposed to the webview.

## Notes

- Base is `main-0.6` (the Holochain 0.6 / 0.20.x release line).
- Ships **lockstep** with the Rust side in holochain/android-service-runtime#137, which injects the env, dispatches `app_request`, and forwards signals.

## Test plan

- `npm run test:tauri` → 22/22 pass (no conductor required); now runs on every PR via the Test workflow.
- Existing websocket e2e tests unaffected.

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -3,8 +3,10 @@ import Emittery from "emittery";
 import IsoWebSocket from "isomorphic-ws";
 import { HolochainError, WsClientOptions } from "./common.js";
 import { AppAuthenticationToken } from "./admin/index.js";
-import { AppSignal, RawSignal, Signal, SignalType } from "./app/index.js";
-import { encodeHashToBase64 } from "../utils/base64.js";
+import {
+  decodeSignal,
+  holoHashMapKeyConverter,
+} from "./app/decode.js";
 
 interface HolochainMessage {
   id: number;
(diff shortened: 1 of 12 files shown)


src/api/client.ts
```diff
@@ -260,30 +262,7 @@ export class WsClient extends Emittery {
...
-            cell_id: encodedAppSignal.cell_id,
-            zome_name: encodedAppSignal.zome_name,
-            payload,
-          };
-          this.emit("signal", {
-            type: SignalType.App,
-            value: signal,
-          } as Signal);
-        }
+        this.emit("signal", decodeSignal(decode(message.data)));
```
jost-s, line 265: Three functions from this file have been moved to `./decode.ts`

jost-s (approved):
LGTM

jost-s:
Should be forward-backported to main

zippy:
Closing this in favor of version off main and will then backport.

PR #461, opened 2026-08-11
feat!: regenerate the TypeScript types from the Holochain source

### LINKED ISSUE #458: Regenerate the TypeScript types from the Holochain source instead of writing them by hand
## Description

The TypeScript types in this repository are written and maintained by hand. They are a copy of the Rust types that the Holochain conductor sends and receives over its WebSocket interfaces. Every time one of those Rust types changes, somebody has to notice it and reproduce the change by hand here. That is slow, easy to forget, and when it is forgotten the two sides drift apart without anyone finding out until an application breaks.

holochain/holochain#5930 makes the Holochain repository able to generate those TypeScript declarations from the Rust source when a dedicated build option is turned on. This issue covers the client side: a script that fetches Holochain and regenerates the types, and a workflow that runs it on every push.

### The npm script

A new script, for example `npm run regen:types`, that does the following.

1. Clones the Holochain repository into a temporary, git ignored directory. The Holochain branch to use is selectable through an argument or an environment variable, and defaults to `develop`.
2. Runs the type export with the dedicated build option turned on, writing the generated files straight into this repository's `src` tree, replacing the hand written type files in place. Which Rust type ends up in which file is defined in holochain/holochain#5930.
3. Runs `npm run lint` so the generated files match the formatting this repository enforces.

Running the script requires a Rust toolchain. Holochain pins the version it needs in its `rust-toolchain.toml`, so the script or its documentation should point people at that.

### The workflow

A new workflow that runs on every push. It runs the script and fails when the regenerated types differ from the committed ones, so drift between the two repositories is caught by the build instead of being discovered by a breaking application.

Since this repository keeps one branch per supported Holochain minor version (`main`, `main-0.6`, `main-0.7` at the time of writing), the workflow has to pass the matching Holochain branch to the script instead of always using the default.

### Preparatory work that has to happen first

These were found while checking the idea against the current code, and each one blocks a working regeneration.

* **The files being overwritten are not purely generated.** `src/api/admin/types.ts` and `src/api/app/types.ts` also hold types that belong to the client and have no counterpart in Rust, for example the request helper types and the WebSocket connection options. Those have to be moved into their own files before anything overwrites the originals, or the client will not compile.
* **The documentation build will reject the generated files.** The documentation tool is configured to treat a missing visibility marker as an error, and every exported type carries a `@public` marker today. The generated types will not have one. Either the generator has to be made to emit it, or the tool has to be reconfigured to allow it. Without this, `npm run build` fails on the very first regeneration.
* **Someone has to check the generated shapes against the current hand written ones.** The two should agree. Where they do not, we need to know whether the client was wrong all along or the generator needs an override, and the first run of this is a good opportunity to find out.

## Dependencies

Blocked by holochain/holochain#5930.

## Notes

Open question: when the workflow finds a difference, should it only fail, or also commit the regenerated files? Failing is the safe default and is what is described above, but the maintainers should have the final say.

## Acceptance criteria

- [ ] An npm script regenerates the TypeScript types from a checkout of Holochain, with the Holochain branch selectable and defaulting to `develop`.
- [ ] The client types that have no Rust counterpart live in their own files and are no longer mixed in with the files that get overwritten.
- [ ] After a regeneration the library still builds, passes its own checks, and the reference documentation builds successfully.
- [ ] A workflow runs the script on every push and fails when the regenerated types differ from the committed ones.
- [ ] The generated shapes have been compared against the current hand written ones, and every difference is either fixed here or resolved with an override in the generator.

veeso:
### Summary

The TypeScript types in this package used to be written and kept up to date by hand, copying whatever the Holochain conductor's Rust code expected. Every time a Rust type changed, someone had to notice and update this repository by hand, which was slow and easy to miss.

This change adds a script, `npm run regen:types`, that generates the TypeScript types directly from the Holochain Rust source instead. It also adds a workflow that runs on every push and fails the build if anyone forgets to regenerate the types after a Holochain change, so the two sides can no longer drift apart silently.

By default the script uses Nix to build Holochain's type exporter against the Holochain revision pinned in this repository, so no separate configuration is needed to regenerate against the normal Holochain release. Someone who wants to generate against a different Holochain branch instead can pass a command line flag or set an environment variable to point at that branch, without touching the pinned revision.

Because the generated types use the same names as the Rust source, a number of existing type and field names changed. This is a breaking change. See CHANGELOG.md for the full list of renames, removed types, and one change that is worth reading closely: the field carrying an app signal's payload was renamed from `payload` to `signal`, and code reading the old field name will now silently receive `undefined` instead of failing to compile.

Client-only types that have no Rust equivalent (helper interfaces, connection options, and so on) were moved into their own files so a future regeneration cannot accidentally overwrite them.

### Files that need a human review

Most of the changed files are generated output and do not need line by line review: everything under `docs/` (regenerated by `npm run build:docs`) and the eleven files listed in `src/.ts-rs-manifest` (`src/api/admin/types.ts`, `src/api/app/types.ts`, `src/types.ts`, and everything under `src/hdk/`). Please focus review on the files that were actually hand-written or hand-edited for this change instead:

- `scripts/regen-types.js`, the new generation script, and `test/regen-types.test.ts`
- `.github/workflows/regen-types.yml`, the workflow that catches drift between the committed types and a fresh regeneration
- `flake.nix` and `flake.lock`, which pin the Holochain revision the types are generated from
- `src/.ts-rs-manifest`, the list the script uses to know which generated files to delete before writing fresh ones
- The client-only files split out so regeneration cannot overwrite them: `src/client-types.ts`, `src/api/admin/client-types.ts`, `src/api/app/client-types.ts`
- The rest of the hand-written client code touched by the rename: `src/api/client.ts`, `src/api/common.ts`, `src/api/index.ts`, `src/api/admin/index.ts`, `src/api/admin/websocket.ts`, `src/api/app/index.ts`, `src/api/app/websocket.ts`, `src/api/app/decode.ts`, `src/api/app/tauri-transport.ts`, `src/environments/launcher.ts`, `src/environments/tauri.ts`, `src/utils/base64.ts`, `src/utils/hash-parts.ts`, `src/utils/holo-hash-map.ts`, `src/index.ts`
- `CHANGELOG.md`, for the full list of renames
- `package.json`, `README.md`, `.github/workflows/test.yml`, and `.github/workflows/release-publish.yml`

### TODO:


.github/workflows/test.yml
```diff
@@ -27,6 +45,17 @@ jobs:
...
+      - name: Cache nix store
+        uses: nix-community/cache-nix-action@v6
+        with:
+          primary-key: >-
+            nix-test-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock') }}
+          restore-prefixes-first-match: nix-test-${{ runner.os }}-
+          purge: true
+          purge-prefixes: nix-test-${{ runner.os }}-
+          purge-created: 0
+          purge-primary-key: never
```
jost-s: Why a second cache next to cachix?

scripts/regen-types.js
```diff
@@ -0,0 +1,122 @@
...
+  const branch = argBranch ?? envBranch;
+  return branch
+    ? [
+        "--override-input",
+        "holochain",
+        `github:holochain/holochain/${branch}`,
+        "--no-write-lock-file",
+      ]
+    : [];
+}
```
jost-s: This wasn't a requirement. Types should be in sync with the Holochain version of the nix flake. If that fails to be read, the type generation should fail.

.github/workflows/regen-types.yml
```diff
@@ -0,0 +1,43 @@
...
+          install_url: https://releases.nixos.org/nix/nix-2.25.4/install
+      - uses: nix-community/cache-nix-action@v6
+        with:
+          primary-key: >-
+            nix-regen-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock') }}
+          restore-prefixes-first-match: nix-regen-${{ runner.os }}-
+          purge: true
+          purge-prefixes: nix-regen-${{ runner.os }}-
+          purge-created: 0
+          purge-primary-key: never
```
jost-s: Also here, why two caches?

flake.nix
```diff
@@ -2,7 +2,14 @@
   description = "Nix shell for Holochain app development";
 
   inputs = {
-    holonix.url = "github:holochain/holonix?ref=main-0.7";
+    holonix.url = "github:holochain/holonix?ref=main";
+    # The holochain revision the generated types in src/ are built from.
+    # `nix flake update holochain` moves it forward.
+    holochain = {
```
jost-s: No, why an extra input when `holonix.inputs.holochain` is already there, and `holonix.inputs.holochain should follow it? Please review Clauded stuff before you ask others to review.

src/.ts-rs-manifest
jost-s, line 1: What's this file for? I don't see it being used anywhere.

CHANGELOG.md
```diff
@@ -4,6 +4,62 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Breaking Changes
```
jost-s: I think that the changelog is written to by the release automation. No manual edits required.

jost-s (changes requested):
Looks largely good to me.

The nix flake change is definitely wrong, and the additional nix caches are a question mark as well.

CHANGELOG.md
veeso: yeah, removed. Also added to CLAUDE.md not to generate it.

flake.nix
veeso: removed the extra holochain

src/.ts-rs-manifest
veeso: it is being used inside of `mergeBindings` inside of regen-types.js. It is used to remove dead files.

So we remove every file not in the manifest, so we can detect easily dead code.

.github/workflows/regen-types.yml
veeso: removed

scripts/regen-types.js
veeso: kept feature, but fixed

.github/workflows/test.yml
veeso: my bad, removed

scripts/regen-types.js
jost-s: What's the use case for this feature?

scripts/regen-types.js
```diff
@@ -0,0 +1,122 @@
...
+    if (entry.isDirectory()) {
+      files.push(...listFiles(join(dir, entry.name), rel));
+    } else {
+      files.push(rel);
+    }
+  }
+  return files.sort();
+}
+
+export function mergeBindings({ srcDir, bindingsDir }) {
```
jost-s: What's the point of keeping a record of the generated types, using a manifest? What speaks against taking the currently present types as the definitive current state, replacing whatever was there before?

scripts/regen-types.js
jost-s: I see now the logic. It's more versatile, but convoluted. We're only adding this feature to Holochain and the JS client repo now, namely on branch `develop`, so it's fine to assume by convention that the feature exists.

It's equally fine to directly use the revision the nix flake points to. It'll be difficult to debug when the flake's rev points to `main-0.7`, but really `develop` is used and types change on `develop`. Or rather we won't notice the drift then.

.github/workflows/regen-types.yml
ThetaSinner, line 1: This shouldn't be its own workflow because that means if can't be part of the `ci_pass` requirement that blocks PR merging.

scripts/regen-types.js
veeso: merging was surely not the best thing, but tracking the files was not wrong.
We have generated types in the same folder of manually implemented code, so we can't just delete the entire folder and regenerate.

I've changed to replace the manifest with a scan of auto-generated file (they have a certain header).

We could accept a breaking change with path to have auto-generated types in a separate module, or have maybe a subfolder for generated types, and a `export *` from there maybe.

EDIT: actually it's doable, changing for this

.github/workflows/regen-types.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+      - uses: cachix/install-nix-action@v31
+        with:
+          install_url: https://releases.nixos.org/nix/nix-2.25.4/install
```
ThetaSinner: Not latest https://nixos.org/download/

.github/workflows/regen-types.yml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: cachix/cachix-action@v17
+        with:
+          name: holochain-ci
+      - uses: actions/setup-node@v7
+        with:
+          node-version: 22
```
ThetaSinner: Two problems with this, the flake should be using Node 24, see https://search.nixos.org/packages?channel=26.05&query=nodejs#show=nodejs and we should be getting node/npm from the flake ideally with a Nix shell.

.github/workflows/test.yml
```diff
@@ -11,6 +11,24 @@ concurrency:
...
+      - name: Install Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+
+      - name: Install NPM packages
+        run: npm ci
+
+      - name: Lint
+        run: npx eslint --ext .ts src test eslint.config.cjs
```
ThetaSinner: This is already defined as a script - https://github.com/holochain/holochain-client-js/blob/main/package.json#L34

scripts/regen-types.js
ThetaSinner: Yeah I still don't get it. The Nix flake would be updated to point at the corresponding branch already. We spawn a Holochain binary in client tests and that comes from the flake, so why not `hc`? We are already set up to test against the right version

scripts/regen-types.js
ThetaSinner, line 1: I'm struggling to see what this script is doing that wouldn't be achieved by `rm -rf ./types && nix develop -c hc generate-ts-types`. 

As it stands, this is a Node script, that requires a Nix build, that copies files when `hc` could be updated to support an `--out-dir` argument, that also forces linting and provides overrides. This can definitely be simplified.

CLAUDE.md
```diff
@@ -0,0 +1,165 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with
+code in this repository.
+
+## What this is
+
+`@holochain/client`: TypeScript client for the Holochain Conductor API
+(Admin API and App API), usable from Node and browsers. Client `0.21.x`
+tracks Holochain `0.7.x`. Published from `lib/` (compiled by `tsc`); `src/`
```
ThetaSinner: This information is in the readme and shouldn't be duplicated

CLAUDE.md
```diff
@@ -0,0 +1,165 @@
...
+  Conventional Commit messages by the release workflow. Do not add entries,
+  fix typos, or touch it in any commit or PR.
+- Releases: run the "Prepare Release" GitHub Actions workflow on `main`,
+  merge the generated PR; the merge publishes to npm.
+- Plans/design notes live in `.claude/plans/` (git-excluded locally); do not
+  commit them.
+- Nix CI jobs use `cachix/cachix-action` (`holochain-ci` cache) only. Don't
+  add `nix-community/cache-nix-action` or similar local nix-store caching
+  next to it; it duplicates what cachix already serves and adds purge
+  logic to maintain for no real speedup.
```
ThetaSinner, line 159: That seems like a rather specific detail, Claude really ought to find that by reading the CI workflows?

flake.nix
```diff
@@ -13,6 +13,17 @@
...
+          # Holochain's TypeScript bindings exporter. Built against
+          # holonix's pinned holochain revision by default; regen-types.js
+          # overrides the `holonix/holochain` input on the command line to
+          # target a different branch (the `ts_rs` feature this binary
+          # needs may not exist on every branch). See "Regenerating the
+          # TypeScript client bindings" in holochain's docs/developer_setup.md.
+          packages.export-ts-bindings = inputs'.holonix.packages.hc.override {
+            cargoExtraArgs =
+              "--features ts_rs,unstable-countersigning --bin export-ts-bindings";
+          };
```
ThetaSinner: It would seem cleaner to define this as a shell package that replaces the default `hc` and then you can just use it like `hc` - rather than having to build the package explicitly to get it.

CLAUDE.md
veeso: it tried to duplicate that before, so I would keep it.

CLAUDE.md
veeso: removed

scripts/regen-types.js
veeso: changed to just that + lint

.github/workflows/test.yml
veeso: done

.github/workflows/regen-types.yml
veeso: done

.github/workflows/regen-types.yml
veeso: updated

.github/workflows/test.yml
```diff
@@ -11,6 +11,50 @@ concurrency:
...
+      - name: Install Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+
+      - name: Install NPM packages
+        run: npm ci
+
+      - name: Lint
+        run: npm run lint:ci
```
jost-s:
```suggestion
      - name: Install NPM packages & lint
        run: nix develop -c $SHELL -c "npm ci && npm run lint:ci"
```

flake.nix
```diff
@@ -4,24 +4,44 @@
...
     holonix.url = "github:holochain/holonix?ref=main-0.7";
     nixpkgs.follows = "holonix/nixpkgs";
+
+    # The `ts_rs`/`export-ts-bindings` feature only exists on Holochain's
+    # `develop` branch, not yet on the `main-0.7` release branch holonix
+    # pins above. Kept as its own locked input (rather than a runtime
+    # --override-input) so it's pinned like everything else and doesn't
+    # touch the holochain revision the dev shell / e2e conductor use.
+    holonix-ts-bindings.url = "github:holochain/holonix?ref=main-0.7";
+    holonix-ts-bindings.inputs.holochain.url =
```
jost-s: Actually I see only now that this branch's flake hasn't been upgraded yet. `holonix` must point to `holonix?ref=main` here, and `export-ts-binding` is already there. Then this override can be removed.

jost-s (changes requested):
Almost there

flake.nix
veeso: fixed

jost-s (dismissed):
All is well

flake.nix
```diff
@@ -11,17 +11,27 @@
...
+      perSystem = { inputs', self', pkgs, ... }:
         {
+          # Holochain's TypeScript bindings exporter. See "Regenerating the
+          # TypeScript client bindings" in holochain's
+          # docs/developer_setup.md.
+          packages.export-ts-bindings =
+            inputs'.holonix.packages.hc.override {
+              cargoExtraArgs =
+                "--features ts_rs,unstable-countersigning --bin export-ts-bindings";
+            };
```
ThetaSinner: I'm only just realizing now that this has been added as a separate binary in the `hc` crate and that's why it isn't just the `hc` CLI that is being built with different features. I don't see a reason it couldn't have been a subcommand in `hc` that is behind a feature flag. That's what I intended when we discussed a custom build of `hc`.

ThetaSinner: This comment got left without a response but it'd have been useful to be told I was wrong on that comment https://github.com/holochain/holochain-client-js/pull/461#discussion_r3802771734

jost-s (dismissed):

ThetaSinner (approved):

